PR #24, opened 2024-02-24
Crosslinker Mover lanthionine

tydingcw:
This is the code to enable lanthipeptide modeling in Rosetta. 
Crosslinker updated to include lanthionine and methyllanthionine connections.
-This includes D/L ALA steriochem for lanthionine and D/L and 3 R/S steriochem for methylanthionine (B(L/D)(R/S))
Dehydrated amino acid patches for DBU and DHU were updated/added
The AmbiguouosNMRDistanceConstraints were updated to work for NCAA (hardcoding was generalized)

Additionally, I noticed that the thioether lariat in simple cyc pep predict only samples cis omega bonds for the N terminal residue. I changed this to trans, which should be energetically preferred.


source/src/protocols/cyclic_peptide/crosslinker/Lanthionine_Helper.cc
```diff
@@ -0,0 +1,311 @@
...
+
+/// @brief Given a pose and a selection of residues, add the linker,
+/// align it crudely to the selected residues, and set up covalent bonds.
+/// @details Must be defined by derived classes.  Version for symmetric poses.
+void
+Lanthionine_Helper::add_linker_symmetric(
+	core::pose::Pose & pose,
+	core::select::residue_selector::ResidueSubset const & selection
+) const {
+	//This will not compile until this is implemented.
```
roccomoretti, line 113: Is this comment still valid? (If not, I suggest removing it and the other similar comments.)

source/src/protocols/cyclic_peptide/crosslinker/Lanthionine_Helper.cc
```diff
@@ -0,0 +1,311 @@
...
+/// @brief Given indices of residues that are already linked to a linker, get the index
+/// of the linker.
+/// @details Returns zero for this class, since no linker is added.
+core::Size
+Lanthionine_Helper::get_linker_index_asymmetric(
+	core::pose::Pose const & /*pose*/,
+	utility::vector1< core::Size > const & /*res_indices*/
+) const {
+	//This will not compile until this is implemented.
+	return 0;
```
roccomoretti, line 169: Does this need to be implemented still?

source/src/protocols/cyclic_peptide/crosslinker/Lanthionine_Helper.cc
```diff
@@ -0,0 +1,311 @@
...
+/// of the indices of the symmetric pieces of the linker.
+/// @details Returns an empty vector for linker indices for this class, since no linker is added.
+void
+Lanthionine_Helper::get_linker_indices_symmetric(
+	core::pose::Pose const & /*pose*/,
+	utility::vector1< core::Size > const & /*res_indices*/,
+	utility::vector1< core::Size > & linker_indices
+) const {
+	//This will not compile until this is implemented.
+	linker_indices.clear();
```
roccomoretti, line 182: Does this need to be implemented still?

source/src/protocols/cyclic_peptide/crosslinker/Lanthionine_Helper.cc
```diff
@@ -0,0 +1,311 @@
...
+			if ( dalares == 0 ) {
+				dalares = i;
+			} else if ( cysres == 0 ) {
+				cysres = i;
+				if ( !pose.residue_type( cysres ).is_sidechain_thiol() ) { cysres=dalares; dalares=i;}
+			} else {
+				utility_exit_with_message( errmsg + "More than two residues were selected!  A lanthionine connection is specified by exactly two residues." );
+			}
+		}
+	}
```
roccomoretti, line 283: The logic here is a bit confused with the multiple swapping opportunities. I would suggest tightening it up by using the get_residues_from_subset() function from core/select/util.hh. This should give you a 2-length vector, and you can then figure out which one is the cys and which one is the dala (only assigning them once.)

source/src/protocols/cyclic_peptide_predict/SimpleCycpepPredictApplication.cc
```diff
@@ -803,6 +813,25 @@ SimpleCycpepPredictApplication::initialize_from_options(
 		}
 	}
 
+	//CWT store lan and melan positions
+	if ( option[basic::options::OptionKeys::cyclic_peptide::lanthionine_positions].user() ) {
+		core::Size const nlanres(option[basic::options::OptionKeys::cyclic_peptide::lanthionine_positions]().size());
+		runtime_assert_string_msg( nlanres > 0, "Error in simple_cycpep_predict application: The \"-cyclic_peptide:lanthionine_positions\" commandline option must be followed by a list of residues to form lanthionine rings." );
+		//runtime_assert_string_msg( nlanres % 2 == 0, "Error in simple_cycpep_predict application: The \"-cyclic_peptide:lanthionine_positions\" commandline option must be followed by a list of residues, where the number of residues in the list is a multiple of two.  Pairs of residues must form lanthionine rings." );
+		//Only allowing two specified residues at the moment
+		runtime_assert_string_msg( nlanres == 2, "Error in simple_cycpep_predict application: The \"-cyclic_peptide:lanthionine_positions\" commandline option must be followed by a list of two residues. Pairs of residues must form lanthionine rings." );
```
roccomoretti, line 822: This seems a bit off. Your loop seems to indicate that you should be able to handle multiple lantionine bonds. Is there a particular reason you're limiting it to only a single bond (two residues) at the moment? You may want to clarify the reason for that here, so the next person has a better sense of what would need to be addressed.

source/src/protocols/cyclic_peptide_predict/SimpleCycpepPredictApplication.cc
```diff
@@ -1761,6 +1791,18 @@ SimpleCycpepPredictApplication::run() const {
...
+			//need to call for all lan and melan resi pairs
+			if ( lanthionine_positions_.size() > 0 ) {
+				for ( core::Size i=1, imax=lanthionine_positions_.size(); i<=imax; ++i ) { //Loop through all sets of doubles of residues.
+					debug_assert(lanthionine_positions_[i].size() == 2); //Should always be true.
+					protocols::cyclic_peptide::crosslinker::set_up_lanthionine_variants(*pose, lanthionine_positions_[i][1], lanthionine_positions_[i][2]);
+				}
+				core::pose::add_lower_terminus_type_to_pose_residue( *pose, 1 );
+				core::pose::add_upper_terminus_type_to_pose_residue( *pose, pose->size() );
+			} else {
+				runtime_assert_string_msg( lanthionine_positions_.size() > 0 , "Error in protocols::cyclic_peptide_predict::SimpleCycpepPredictApplication lanthionine positions are empty");
```
roccomoretti, line 1804: I'll mention that you probably don't need to have an explicit `if()` test and then runtime_assert with the same condition. If the assumption is that you'll only reach this point if you have a valid  lanthionine_positions_ setting, you can just do the runtime_assert, and continue with the "passed" code at the same level.

source/src/protocols/cyclic_peptide_predict/SimpleCycpepPredictApplication.cc
```diff
@@ -803,6 +813,25 @@ SimpleCycpepPredictApplication::initialize_from_options(
...
+		core::Size const nlanres(option[basic::options::OptionKeys::cyclic_peptide::lanthionine_positions]().size());
+		runtime_assert_string_msg( nlanres > 0, "Error in simple_cycpep_predict application: The \"-cyclic_peptide:lanthionine_positions\" commandline option must be followed by a list of residues to form lanthionine rings." );
+		//runtime_assert_string_msg( nlanres % 2 == 0, "Error in simple_cycpep_predict application: The \"-cyclic_peptide:lanthionine_positions\" commandline option must be followed by a list of residues, where the number of residues in the list is a multiple of two.  Pairs of residues must form lanthionine rings." );
+		//Only allowing two specified residues at the moment
+		runtime_assert_string_msg( nlanres == 2, "Error in simple_cycpep_predict application: The \"-cyclic_peptide:lanthionine_positions\" commandline option must be followed by a list of two residues. Pairs of residues must form lanthionine rings." );
+		core::Size count(0);
+		lanthionine_positions_.resize(nlanres / 2);
+		for ( core::Size i(1), imax(nlanres / 2); i<=imax; ++i ) {
+			utility::vector1 <core::Size> innervect(2);
```
roccomoretti, line 826: vector1 works, but if you know you'll always have just two elements, std::pair is likely to be a better choice. (But no need to re-write things just because.)

source/src/protocols/cyclic_peptide_predict/SimpleCycpepPredictApplication.cc
```diff
@@ -2160,6 +2213,39 @@ SimpleCycpepPredictApplication::find_first_and_last_thioether_lariat_residues(
 	}
 }
 
+/// @brief Given a pose, find the first and last lanthipeptide bond-forming residues.
+/// @details Randomly choose based on input residue numbers
+void
+SimpleCycpepPredictApplication::find_first_and_last_lanthipeptide_residues(
```
roccomoretti, line 2219: Slight nomenclature note: "find" doesn't really mesh with random choice in my mind (finding tends to lean more toward the deterministic search). I would personally go with "select" or something like that.

source/src/protocols/cyclic_peptide_predict/SimpleCycpepPredictApplication.cc
```diff
@@ -3232,6 +3374,17 @@ SimpleCycpepPredictApplication::add_closebond_logic_thioether_lariat(
...
 	// In theory you could randomize CONN2-CP2-CO-N and CO-N-CA-C but that's tomorrow's problem. AMW TODO
+
+	//This application appears to only produce cis CA-N-CO-CP2 bonds without randomization,
+	// trans should be energetically favored, so setting bond to trans here
+	genkic->add_perturber( protocols::generalized_kinematic_closure::perturber::set_dihedral );
+	utility::vector1< core::id::NamedAtomID > omega_vect;
+	omega_vect.push_back( core::id::NamedAtomID( "N", cyclization_point_end));
+	omega_vect.push_back( core::id::NamedAtomID( "CO", cyclization_point_end));
+	genkic->add_atomset_to_perturber_atomset_list( omega_vect );
+	genkic->add_value_to_perturber_value_list( 180.0 );
```
roccomoretti, line 3385: Does this addition change the behavior of any current protocols? Is that acceptable in non-lanthionine cases?

source/src/protocols/cyclic_peptide_predict/SimpleCycpepPredictApplication.cc
```diff
@@ -3286,6 +3439,79 @@ SimpleCycpepPredictApplication::add_closebond_logic_thioether_lariat(
...
+SimpleCycpepPredictApplication::add_closebond_logic_lanthipeptide(
+	core::pose::PoseCOP pose,
+	core::Size const cyclization_point_start,
+	core::Size const cyclization_point_end,
+	protocols::generalized_kinematic_closure::GeneralizedKICOP genkic
+) const {
+	debug_assert( cyclization_type() == SCPA_lanthipeptide );
+	core::Size ala_res = 0;
+	core::Size cys_res = 0;
+	if ( pose->residue_type(cyclization_point_start).base_name() == "CYS" ) {
```
roccomoretti, line 3453: Is the deliberate comparison to CYS intended, or is generlization to sidechain thiol a better choice here?

source/src/protocols/cyclic_peptide/crosslinker/lanthionine_util.cc
```diff
@@ -0,0 +1,620 @@
...
+	core::chemical::ResidueType const & restype( pose.residue_type( dalares ) );
+	core::Size const dalares_sc_connection_id( restype.n_possible_residue_connections() ); //We will assume that the highest-numbered residue connection is the sidechain connection ID.
+	core::Size const dalares_sc_connection_atom_index( restype.residue_connect_atom_index( dalares_sc_connection_id ) );
+	std::string const dalares_sc_connection_atom( restype.atom_name( dalares_sc_connection_atom_index ) );
+
+	//Get the name of the second sidechain connection atom:
+	core::chemical::ResidueType const & restype2( pose.residue_type( cysres ) );
+	core::Size const cysres_sc_connection_id( restype2.n_possible_residue_connections() ); //We will assume that the highest-numbered residue connection is the sidechain connection ID.
+	core::Size const cysres_sc_connection_atom_index( restype2.residue_connect_atom_index( cysres_sc_connection_id ) );
+	std::string const cysres_sc_connection_atom( restype2.atom_name( cysres_sc_connection_atom_index ) );
```
roccomoretti, line 115: Would it be worth checking this with the disulfide_atom_name() results?

roccomoretti (commented):
Looks pretty good. Just some various comments. Nothing much that's strictly necessary.

I would double check the results of the `autoNOE_rosetta and the noe_assignment integration tests -- the changes might just be cosmetic, but you'll want to pick through the changes to make sure nothing has changed functionally.

The simple_cycpep_predict_* integration tests results is a bit more concerning -- do you know why there's a change in the structures they produce? You should understand the changes prior to merging.

source/src/protocols/cyclic_peptide_predict/SimpleCycpepPredictApplication.cc
tydingcw: This changes the behavior of the thioether lariat. It was predicting the N terminal peptide bond to always be cis. This changes it to be trans, which causes the test difference. I messaged @vmullig about this but have not heard back yet.

roccomoretti: Do we have good experimental data on the cis/trans prevalence in thioether lariats?

source/src/protocols/cyclic_peptide/crosslinker/Lanthionine_Helper.cc
tydingcw: Done

source/src/protocols/cyclic_peptide/crosslinker/Lanthionine_Helper.cc
tydingcw: Done

source/src/protocols/cyclic_peptide/crosslinker/Lanthionine_Helper.cc
tydingcw: Done

source/src/protocols/cyclic_peptide_predict/SimpleCycpepPredictApplication.cc
tydingcw: The reason is that SimpleCycpepPredict is not currently able to adjacent rings or overlapping rings. I added a comment in the code to address this.

source/src/protocols/cyclic_peptide_predict/SimpleCycpepPredictApplication.cc
tydingcw: This keeps the code consistent with other functions in the application.

source/src/protocols/cyclic_peptide_predict/SimpleCycpepPredictApplication.cc
tydingcw: I am not sure how much data there is. I know that in this structure (https://www.rcsb.org/structure/8FEG), the thioether is conjugated to a trans peptide bond.

source/src/protocols/cyclic_peptide_predict/SimpleCycpepPredictApplication.cc
tydingcw: For lanthipeptides, the mechanism of formation involves an L-cysteine residue.

source/src/core/scoring/constraints/AmbiguousNMRDistanceConstraint.cc
```diff
@@ -74,6 +76,49 @@ bool is_aromatic( core::chemical::AA aa ) {
...
+		} else if ( name.substr(0, 1) == "Q" ) {
+			name = 'H' + name.substr(1, name.size() - 1);
+		}
+
+		//Now check for all atom names containing the substring
+		for ( core::Size num = 1; num <= pose.residue_type(res).natoms(); ++num ) {
+			//std::cout << "Num: " << num << " of" << pose.residue_type(res).natoms() << std::endl;
+			size_t find = pose.residue_type(res).atom_name(num).find(name);
+			if ( find != std::string::npos ) {
+				auto atm_name = pose.residue_type(res).atom_name(num);
```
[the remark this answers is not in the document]
tydingcw: Done

tests/integration/tests/crosslinkermover_lanthionine/flags
```diff
@@ -0,0 +1,9 @@
+-parser:protocol inputs/test.xml
+-in:file:fasta inputs/dummy.fasta
+-nstruct 10
```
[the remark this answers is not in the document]
tydingcw: This was changed to 5, but want some backups in case of ring closure failure.

source/src/protocols/cyclic_peptide_predict/SimpleCycpepPredictApplication.hh
```diff
@@ -13,6 +13,8 @@
 /// using generalized kinematic closure (GenKIC) for cyclization, and enforcing user-defined requiresments for numbers of mainchain hydrogen bonds.
 /// @author Vikram K. Mulligan, Baker laboratory (vmullig@uw.edu)
 
+//CTW TODO add lanthipeptides
+
```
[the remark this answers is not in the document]
tydingcw: Removed

source/src/protocols/cyclic_peptide_predict/SimpleCycpepPredictApplication.cc
roccomoretti: I'm just thinking about future engineering applications. Would there (could there) ever be the case where you might want to model a lanthipeptide with some non-canonical amino acid which isn't starting with an L-cysteine?

source/src/protocols/cyclic_peptide/crosslinker/lanthionine_util.cc
tydingcw: I am unsure. This code is all modeled off the thioether_util.cc code.

source/src/core/scoring/constraints/AmbiguousNMRDistanceConstraint.cc
```diff
@@ -74,6 +76,49 @@ bool is_aromatic( core::chemical::AA aa ) {
...
+				atm_name.erase(std::remove_if(atm_name.begin(), atm_name.end(), ::isspace), atm_name.end());
```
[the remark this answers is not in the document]
tydingcw: I changed this to use the `stripped_whitespace()` function

roccomoretti (approved):

PR #26, opened 2024-03-12
QOL update to the PerResidueEsmProbabilitiesMetric

MoritzErtelt:
This short PR:
1. Adds the proper citation for ESM
2. Adds some checks for NaNs/Infs to the Perplexity/softmax calculation
3. fixes a typo in MIFST.cc
4. Prevents an error in SampleSequenceFromProbabilities when calling the SimpleThreadingMover without mutations
5. Fixes const behavior in SampleSequenceFromProbabilities


source/src/protocols/esm_perplexity/PseudoPerplexityMetric.cc
```diff
@@ -124,8 +124,11 @@ PseudoPerplexityMetric::compute_perplexity(
 		}
 		core::Real aa_probability = position_pair.second.at( aa_type );
 		// if the probability is zero add a small constant to it to prevent -inf
-		if ( aa_probability == 0 ) {
-			aa_probability += std::numeric_limits< core::Real >::min();
+		if ( aa_probability == 0 || aa_probability < 0.00001 ) {
+			aa_probability += 0.00001;
+		}
+		if ( std::isnan(aa_probability) || std::isinf(aa_probability)) {
+			aa_probability += 0.00001;
```
roccomoretti, line 131: I don't know what this is doing. Adding anything to inf or nan will still result in inf and nan.

source/src/protocols/simple_moves/SampleSequenceFromProbabilities.cc
```diff
@@ -431,7 +443,7 @@ SampleSequenceFromProbabilities::calculate_position_diffs(
 		}
 
 		core::Real max_prob = 0.0;
-		for ( auto aa_and_prob : aa_probs ) {
```
roccomoretti, line 434: Is this why you remove the const from the function signature? You should be able to do `auto const aa_and_prob` and maintain the function signature.

roccomoretti (commented):
Some comments, and you need to beautify. (It looks like the beautify function on the server isn't working, probably because we haven't figured out how to push to people's forks.

Just try running `python ../tools/python_cc_reader/beautify_changed_files_in_branch.py` from the `rosetta/source/` directory instead.

source/src/protocols/esm_perplexity/PseudoPerplexityMetric.cc
MoritzErtelt: Ah yes that's wrong, I changed it to "=" as I intended instead of "+=", thanks for the catch!

source/src/protocols/simple_moves/SampleSequenceFromProbabilities.cc
MoritzErtelt: No that's not the reason, basically in the `SamplingSequenceFromProbabilities` mover I am copying the probabilities stored in the given metric once, and then pass these around sometimes modifying them. E.g. if a user has specified a task operation to restrict design, I set the probabilities of other AAs for that residue to zero. Just had a slight mixup in the previous version that is now fixed.

MoritzErtelt:
Thank you for the review Rocco, I've beautified locally and will let the test server have another go at it.

roccomoretti (approved):

PR #27, opened 2024-03-14
Update README.md: git clone --recursive

smoe:
The way I did it was to first clone and then initialize+retrieve the submodules afterwards - this is what I should have done and I guess others will appreciate that approach, too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,9 @@ Start here: https://www.rosettacommons.org/docs/latest/getting_started/Getting-S
 The fastest way to address your scientific modeling challenge at hand may be with the official Docker images.
 Accessible from the official Docker hub at https://hub.docker.com/r/rosettacommons/rosetta,
 the images have both Rosetta and PyRosetta pre-installed, such that Rosetta tutorials can be followed.
+You may also conveniently inspect the Dockerfiles
+(see [docker/README.md](https://github.com/RosettaCommons/rosetta/blob/main/docker/README.md))
+as a templated for the Rosetta installation on your local system.
 
 Anticipating that you like what you see or if coming with a Computer Science background,
 you may decide to extend and compile the Rosetta source tree yourself. See the
--- a/README.md
+++ b/README.md
@@ -56,6 +56,15 @@ Some C++ compilers may not yet perfectly master these features.
 Later versions of those compilers have that fixed.
 Talk back to us when you run into issues.
 
+Docker
+======
+
+Official Rosetta/PyRosetta images could be found at https://hub.docker.com/r/rosettacommons/rosetta.
+Both `serial` and `mpi` Rosetta builds provided as well as the number of PyRosetta builds including fully functional Jupyter setups with PyRosetta pre-installed and experimenta builds with `libtorch` and `tensorflow` integration.
+Please see https://hub.docker.com/r/rosettacommons/rosetta for more information.
+
+Various reference Docker files could be found in `rosetta/docker` dir.
+
 PyRosetta
 =========
 
@@ -63,28 +72,19 @@ PyRosetta are Python bindings to the Rosetta library. These can be built from th
 
 See <https://www.pyrosetta.org> for more information about PyRosetta.
 
-Docker
-======
-
-Official Rosetta/PyRosetta images could be found at https://hub.docker.com/r/rosettacommons/rosetta. Both `serial` and `mpi` Rosetta builds provided as well as the number of PyRosetta builds including fully functional Jupyter setups with PyRosetta pre-installed and experimenta builds with `libtorch` and `tensorflow` integration. Please see https://hub.docker.com/r/rosettacommons/rosetta for more information.
-
-Various reference Docker files could be found in `rosetta/docker` dir.
-
-
 The Docker image referenced above already ships with PyRosetta.
-To prepare the PyRosetta Python module locally from this source tree, do the following:
+To prepare the PyRosetta Python module locally from this source tree,
+you need
+ * a C++ compiler, like the one you used to compile the other parts of Rosetta
+ * the Ninja ([conda](https://anaconda.org/conda-forge/ninja), [Debian](https://tracker.debian.org/pkg/ninja-build)) build management tool 
+ * and also CMake ([conda](https://anaconda.org/conda-forge/cmake), [Debian](https://tracker.debian.org/pkg/cmake))
+which should all be readily available from your regular Linux distribution.
 
 ``` sh
 $ cd source/src/python/PyRosetta
-$ cowsay "Abracadabra!"
- ______________
-< Abracadabra! >
- --------------
-        \   ^__^
-         \  (oo)\_______
-            (__)\       )\/\
-                ||----w |
-                ||     ||
+$ python3 build.py -j24 --create-package $HOME/my_pyrosetta_package
+$ cd $HOME/my_pyrosetta_package/setup
+$ python3 setup.py install
 ```
 
 Developing Rosetta
--- a/README.md
+++ b/README.md
@@ -3,7 +3,8 @@ Rosetta Biomolecular Modeling Library
 
 The Rosetta software suite includes algorithms for computational modeling and analysis of protein structures. It has enabled notable scientific advances in computational biology, including de novo protein design, enzyme design, ligand docking, and structure prediction of biological macromolecules and macromolecular complexes.
 
-Rosetta is maintained by the RosettaCommons, a collaboration of 50+ academic research groups, who have been developing Rosetta for over 20 years. See <https://ww.rosettacommons.org> for more information about Rosetta and the RosettaCommons.
+Rosetta is maintained by the RosettaCommons, a collaboration of 50+ academic research groups, who have been developing Rosetta for over 20 years.
+See <https://www.rosettacommons.org> for more information about Rosetta and the RosettaCommons.
 
 Rosetta Code
 ============
@@ -76,7 +77,7 @@ The Docker image referenced above already ships with PyRosetta.
 To prepare the PyRosetta Python module locally from this source tree,
 you need
  * a C++ compiler, like the one you used to compile the other parts of Rosetta
- * the Ninja ([conda](https://anaconda.org/conda-forge/ninja), [Debian](https://tracker.debian.org/pkg/ninja-build)) build management tool 
+ * the Ninja ([conda](https://anaconda.org/conda-forge/ninja), [Debian](https://tracker.debian.org/pkg/ninja-build)) build management tool
  * and also CMake ([conda](https://anaconda.org/conda-forge/cmake), [Debian](https://tracker.debian.org/pkg/cmake))
 which should all be readily available from your regular Linux distribution.
 
@@ -106,7 +107,7 @@ The main repository (RosettaCommons/rosetta) contains the Rosetta source code, d
 * rosetta/source/src -- The Rosetta source
 * rosetta/database/ -- The Rosetta database (used during runtime)
 * rosetta/source/test/ -- The compiled unit tests
-* rosetta/test/integration/ -- The integration tests
+* rosetta/tests/integration/ -- The integration tests
 * rosetta/source/bin/ -- The location of the (symlinks to) the Rosetta executables
 * rosetta/source/build/ -- The location of the built libraries
 
--- a/README.md
+++ b/README.md
@@ -12,12 +12,11 @@ Rosetta Code
 While the Rosetta source code is published on GitHub, it is not "Open Source" (according to the OSI definition). Most notably, use for commercial purposes requires purchase of a separate license. See LICENSE.md for further information.
 
 The main GitHub repository on https://github.com/RosettaCommons/rosetta integrates all the Rosetta-associated code base.
-It should be noted that many parts of Rosetta are structured as GitHub projects on their own, which the main repository conveniently presents as submodules.
-Dominant are external software packages that Rosetta is redistributing, which will help synchronizing development to exact versions of those external softwares across software distributions
+It should be noted that many parts of Rosetta are structured as separate GitHub repositories, which the main repository conveniently presents as submodules.
+A significant number are external software packages that Rosetta is redistributing, which helps synchronize development to exact versions of those external softwares across software distributions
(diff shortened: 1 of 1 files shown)


roccomoretti:
Sounds mostly good, at least for Rosetta users. (I will note that submodules necessary for compilation will be automatically downloaded/checked out by the compilation process, so you only need to pull down the submodules if you need the extras in those directories.)

The one caveat is that our experience has shown that for people _developing_ Rosetta, automatically checking out all submodules tends to complicate things, as every time you merge or pull you need to remember to update the submodules, otherwise the on-disk submodule state gets stale and you risk committing an errant version of the submodule.

A similar issue (to a lesser extent) applies to people just using Rosetta, as if they ever change the version of Rosetta they use, their submodule state can lag behind, and would need to re-update the submodules for the new checked out version.

smoe:
As a user, I feel that using Rosetta feels darn close to developing  some corner of it, such that I want to propose not to make any difference between users and developers when it comes to the preferred command(s) to check out the source tree. Also, the bug/problem reports will become much better to interpret when all your versions are identical. So, following
https://stackoverflow.com/questions/4611512/is-there-a-way-to-make-git-pull-automatically-update-submodules
maybe
```
git config --global submodule.recurse true
```
should be added?

I admit that in the meantime I became a bit unhappy about having all those external dependencies when Debian/Ubuntu feature quite some of those already - at some point in the past I had personally taken care that all the build-deps of Rosetta became part of Debian and part of Ubuntu with it - I need to revisit that, though. And of course my local compilation hangs in one of those redundant-to-Debian external source trees. With that inconvenience in mind, maybe the idea to check out all the submodules as proposed in this patch is a bad one? If I understood you right, then "<your distro's package manager> install ..." followed by the invocation of scons.py you would prefer?

roccomoretti:
Ah, that git config seems like a potentially useful settings. (It's just since git 2.15, which may be why I missed it.) I wouldn't necessarily suggest `--global`, as that would affect all repos on the system, not just Rosetta, but recommending the setting might be a nice idea for casual users.

While I agree that in practice it might be nice to have all the dependencies be installed separately (rather than as part of Rosetta), dependency hell is a thing, and experience has taught us that minimizing the externally provided dependencies makes life easier all around, so we lean toward providing the exact versions we need, in known paths. (This is particularly valuable on HPC clusters, where getting the sysadmins to install packages, especially to install them in locations compatible with the build system, can be difficult.) -- I actually don't think we have any support currently for providing the dependencies fully externally. The build system is pretty well hard-coded to expect things from within the Rosetta environment, rather than the distro package manager provided path.

lyskov:
@smoe i want to also add that one issue with cloning all submodules by default will be potentially large disk usage. So for modules that is optional for compilation it might be best to clone them on the need-to-have basis.

smoe:
Well, yes, prior to compilation:
``` sh
$ du -sh .
 23G
 ```
That is a lot, but it is a smaller fraction of my "disk" than the fraction of my time that would now go into Rosetta, so I am not ultimately concerned. :-)

For Linux users, should those now checking out this repository possibly copy the settings of the build servers? Are those setups available for inspection? And for those and/or the Mac users, I recall you have developed a conda recipe - maybe that could make it into the source tree for us mere mortals to follow?

**Comment by smoe** [comment]
I just checked out the source tree again and just ran scons.py without attending the missing submodules myself - and it just magically worked. That was truly unexpected for me. Scons performed submodule cloning for me and it all needed some 2GB, not 25 :) I have now rephrased that README accordingly.

**Comment by lyskov** [comment]
> For Linux users, should those now checking out this repository possibly copy the settings of the build servers? Are those setups available for inspection? And for those and/or the Mac users, I recall you have developed a conda recipe - maybe that could make it into the source tree for us mere mortals to follow?

-- I am planning to publish reference Docker files which could be used as starting points for setting things up. However i must say that we really try to avoid requiring anything special: any modern distribution with compiler chain installs and zlib should be able to compile and run Rosetta/PyRosetta.

-- also note that we now have official Docker images with Rosetta/PyRosetta pre-installed, see https://hub.docker.com/r/rosettacommons/rosetta - i will update our repo readme to reflect this in the next few days.

**Comment by smoe** [comment]
> > For Linux users, should those now checking out this repository possibly copy the settings of the build servers? Are those setups available for inspection? And for those and/or the Mac users, I recall you have developed a conda recipe - maybe that could make it into the source tree for us mere mortals to follow?
> 
> -- I am planning to publish reference Docker files which could be used as starting points for setting things up.

Just to avoid any misunderstanding, I also (?) kind of dislike to have a demand for Docker as a prerequisite for the compilation. I just thought that it might be the closest that I could get to something formal on how to properly compile Rosetta (and PyRosetta). I found the linux-anvil directory but just as the README said, It is only gettting conda up and running on Fedora.

> However i must say that we really try to avoid requiring anything special: any modern distribution with compiler chain installs and zlib should be able to compile and run Rosetta/PyRosetta.

That is truly great to hear! Many thanks!
IMHO the Docker images are a perfect means to learn if an apparent problem is reproducible in a reference build, but they do the Rosetta Community a disservice by introducing an artificial barrier.

> -- also note that we now have official Docker images with Rosetta/PyRosetta pre-installed, see https://hub.docker.com/r/rosettacommons/rosetta - i will update our repo readme to reflect this in the next few days.

I looked at https://hub.docker.com/r/rosettacommons/rosetta/tags but did not find anything I could use to direct my local build.

On a sidenote - my PyRosetta build needed to link against all the Rosetta-provided libraries (including libxml2) and fails over that:

```
....
No changes in include files detected, skipping Binder run...
No changes in source files detected, skipping CMake run...
Building...
cd /mariner/home/moeller/Github/rosetta/source/build/PyRosetta/Linux-5.15.0-2-amd64-x86_64-with-glibc2.37/clang-version/python-3.11/release/build/ && ninja -j1
[1/1] Linking CXX shared library pyrosetta/rosetta.so
FAILED: pyrosetta/rosetta.so
: && /usr/bin/clang++ -fPIC -std=c++11 -fvisibility=hidden -fvisibility-inlines-hidden -O3 -DNDEBUG  -Wl,-rpath,\$ORIGIN/lib -shared -Wl,-soname,rosetta.so -o pyrosetta/rosetta.so @CMakeFiles/rosetta.rsp
  && cd /mariner/home/moeller/Github/rosetta/source/build/PyRosetta/Linux-5.15.0-2-amd64-x86_64-with-glibc2.37/clang-version/python-3.11/release/build && strip /mariner/home/moeller/Github/rosetta/source
/build/PyRosetta/Linux-5.15.0-2-amd64-x86_64-with-glibc2.37/clang-version/python-3.11/release/build/pyrosetta/rosetta.so
/usr/bin/ld: cannot find -lprotocols.8: No such file or directory
/usr/bin/ld: cannot find -lprotocols.7: No such file or directory
/usr/bin/ld: cannot find -lprotocols_e.6: No such file or directory   
/usr/bin/ld: cannot find -lprotocols_d.6: No such file or directory   
/usr/bin/ld: cannot find -lprotocols_c.6: No such file or directory   
/usr/bin/ld: cannot find -lprotocols_b.6: No such file or directory   
/usr/bin/ld: cannot find -lprotocols_a.6: No such file or directory   
/usr/bin/ld: cannot find -lprotocols_h.5: No such file or directory   
/usr/bin/ld: cannot find -lprotocols_g.5: No such file or directory   
/usr/bin/ld: cannot find -lprotocols_f.5: No such file or directory   
/usr/bin/ld: cannot find -lprotocols_e.5: No such file or directory   
/usr/bin/ld: cannot find -lprotocols_d.5: No such file or directory   
/usr/bin/ld: cannot find -lprotocols_c.5: No such file or directory   
/usr/bin/ld: cannot find -lprotocols_b.5: No such file or directory   
/usr/bin/ld: cannot find -lprotocols_a.5: No such file or directory   
/usr/bin/ld: cannot find -lprotocols.4: No such file or directory
/usr/bin/ld: cannot find -lprotocols.3: No such file or directory
/usr/bin/ld: cannot find -lprotocols_b.2: No such file or directory   
/usr/bin/ld: cannot find -lprotocols_a.2: No such file or directory   
/usr/bin/ld: cannot find -lprotocols.1: No such file or directory
/usr/bin/ld: cannot find -lcore.6: No such file or directory
/usr/bin/ld: cannot find -lcore.5: No such file or directory
/usr/bin/ld: cannot find -lcore.4: No such file or directory
/usr/bin/ld: cannot find -lcore.3: No such file or directory
/usr/bin/ld: cannot find -lcore.2: No such file or directory
/usr/bin/ld: cannot find -lcore.1: No such file or directory
/usr/bin/ld: cannot find -lbasic: No such file or directory
/usr/bin/ld: cannot find -lnumeric: No such file or directory
/usr/bin/ld: cannot find -lutility: No such file or directory
/usr/bin/ld: cannot find -lObjexxFCL: No such file or directory
/usr/bin/ld: cannot find -lcppdb: No such file or directory
/usr/bin/ld: cannot find -lbcl: No such file or directory
/usr/bin/ld: cannot find -llibzmq: No such file or directory
/usr/bin/ld: cannot find -llibxml2: No such file or directory
/usr/bin/ld: note to link with /lib/x86_64-linux-gnu/libxml2.a use -l:libxml2.a or rename it to liblibxml2.a
/usr/bin/ld: cannot find -lrdkit: No such file or directory
/usr/bin/ld: cannot find -lcmaes: No such file or directory
/usr/bin/ld: cannot find -lcifparse: No such file or directory
clang: error: linker command failed with exit code 1 (use -v to see invocation)
ninja: build stopped: subcommand failed.
```
I have successfully compiled PyRosetta in the past following instructions from the PyRosetta website, now I had checked out the PyRosetta submodules and found another README that I followed.  I suggest to extend the toplevel README also with an instruction on how to compile PyRosetta, since those libraries are available, like:
```
$ find . -name "librd*"
./source/build/external/debug/linux/5.15/64/x86/gcc/13/default/librdkit.so
./source/build/external/debug/linux/5.15/64/x86/clang/16.0/default/librdkit.so
```
Thank you tons also for the upcoming update of the README. Hoping to help, I copied the references you put in this thread to my PR - not to get the PR accepted, just to help exchange what new users may need to become productive (and this I mean in two ways - productive for the users' problems and also productive to improve the Rosetta code base).

**Comment by lyskov** [comment]
re PyRosetta compilation: - could it be that you referring to https://github.com/RosettaCommons/rosetta/blob/main/source/src/python/PyRosetta/README.md ? - this mode of compilation is no longer supported and i am planning to update/remove it in the next few days. For now please use current build instructions outlined here: https://www.pyrosetta.org/documentation#h.q91xbiqnd5yv

**Comment by smoe** [comment]
The README.md now integrates the build instructions from the (once familiar) website you referenced. Still building on my end :)
Have rebased on main.

I locally had a different hash for the binder submodule, which by some weird git-ways had made it into an earlier local version of my HEAD. This seems to be fixed but please kindly double-check, just in case.

**Comment by smoe** [comment]
I can now happily report that these instructions worked for me:
```
...
[8393/8399] Building CXX object CMakeFiles/rosetta.dir/protocols/vip/VIP_Mover.cpp.o
[8394/8399] Building CXX object CMakeFiles/rosetta.dir/utility/io/FileContentsMap.cpp.o
[8395/8399] Building CXX object CMakeFiles/rosetta.dir/utility/version.cpp.o
[8396/8399] Building CXX object CMakeFiles/rosetta.dir/utility/xsd_util/util.cpp.o
[8397/8399] Building CXX object CMakeFiles/rosetta.dir/utility/signals/BufferedSignalHub_3.cpp.o
[8398/8399] Building CXX object CMakeFiles/rosetta.dir/core/fragment/FragCache.cpp.o
[8399/8399] Linking CXX shared library pyrosetta/rosetta.so

Creating Python package at: /home/moeller/my_pyrosetta_package...
```

File: README.md [hunk 1]
```diff
@@ -8,34 +8,83 @@ Rosetta is maintained by the RosettaCommons, a collaboration of 50+ academic res
...
+ * a C++ compiler, like the one you used to compile the other parts of Rosetta
+ * the Ninja ([conda](https://anaconda.org/conda-forge/ninja), [Debian](https://tracker.debian.org/pkg/ninja-build)) build management tool 
+ * and also CMake ([conda](https://anaconda.org/conda-forge/cmake), [Debian](https://tracker.debian.org/pkg/cmake))
+which should all be readily available from your regular Linux distribution.
+
+``` sh
+$ cd source/src/python/PyRosetta
+$ python3 build.py -j24 --create-package $HOME/my_pyrosetta_package
+$ cd $HOME/my_pyrosetta_package/setup
+$ python3 setup.py install
```
lyskov, line 80: I have related PR https://github.com/RosettaCommons/rosetta/pull/41/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R32 which try to fix the same things. 

@roccomoretti what is your take on this? One way to do this would be to merge source/src/python/PyRosetta/README.md and this README files or we can keep PyRosetta readme separate at source/src/python/PyRosetta/README.md - thoughts?

smoe: I was not asked, please forgive: Since PyRosetta is a project on GitHub in its own right, and likely so for more than only historic reasons, IMHO it needs a README that can stand for itself.  I have not yet thought too much over what the PyRosetta README I would expect to present, but if you consider this exchange over this README.md fruitful then I would come up with a likely similar PR also for the PyRosetta README.md.

smoe:
The README still lacks two things that I now have not yet added and (for the second point) do not really know what to add:
 1. Pointers to the Dockerfiles to direct the preparation of the local Linux installation fo Rosetta.
 2. Clarification what documentation/tutorial to follow. The GitHub community is likely to find the https://github.com/RosettaCommons/PyRosetta.notebooks as their first entry, which I find inferior to https://new.rosettacommons.org/docs/latest .

smoe:
Rebased to main after merge of #41 . Something I had noted early but kind of brushed aside is that "Getting started" starts with how to build the source tree. I suggest to introduce a "How to build" subsection and have "Getting started" refer to tutorials, documentation and explain build vs docker.

roccomoretti (approved):
I made some adjustments to phrasing and organization, and I added in a few things which were discussed but not quite handled, but I think this is good to go now.

PR #32, opened 2024-03-15
add support for PyRosetta ML builds

lyskov:
add `--torch` option

### DIFF
--- a/source/src/protocols/inverse_folding/MIFST.hh
+++ b/source/src/protocols/inverse_folding/MIFST.hh
@@ -38,6 +38,9 @@
 #include <utility/vector1.hh>
 
 // Torch header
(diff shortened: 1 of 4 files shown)


weitzner:
Do we want to do this as opposed to supporting importing torch and tensorflow the way we do for hdf5?

lyskov:
@weitzner i do not think it one-vs-the-other, - in perfect world we can have both. I do agree that in general we should prefer Python implementation but: this could be made available *now* while so far no one volunteered to rewrite corresponding movers in Python 🤷🏻‍♂️.

weitzner:
My concern is about the long-term maintainability of this solution as well as the size of the build artifacts.

PR #33, opened 2024-03-16
Support python bitwise operators for residue selectors in PyRosetta

klimaj:
This PR adds support for use of python bitwise operators for `ResidueSelector` objects in PyRosetta. Residue selectors are now able to be combined by the following PyRosetta-to-python operator symbol conversions:

- `AndResidueSelector` --> `&`
- `OrResidueSelector` --> `|`
- `NotResidueSelector` --> `~`
- Exclusive `OrResidueSelector` --> `^` (there isn't technically an exclusive `OrResidueSelector` in PyRosetta, but it could be implemented with the `Logic` residue selector in XML. This is intended to be equivalent to a XOR operation or the symmetric difference in set theory)
- In-place `AndResidueSelector` --> `&=`
- In-place `OrResidueSelector` --> `|=`
- In-place exclusive `OrResidueSelector` --> `^=`

Only python bitwise operators are affected, while python boolean operators `and`, `or` and `not` are unaffected when implemented with `ResidueSelector` objects. 

Also thanks to @everyday847 for the original idea to override the `__and__` method of residue selector objects.


lyskov (changes requested):
+1 on this idea @klimaj ! Minor comment: looks like you added test for it as another module for distributed module which does not sounds right to me. Would you be open to the idea of moving test file (ie `test_residue_selectors.py`) into https://github.com/RosettaCommons/rosetta/tree/main/source/src/python/PyRosetta/src/test dir named something like `T121_core.residue-selectors.py`? That way it could be treated as test suite and run in parallel. Thanks,

klimaj:
Thanks @lyskov, that's a good point -- I've now moved the unit test to `src/test/T121_core.ResidueSelector.py` and it's passing on the benchmark server!

lyskov (approved):
LGTM!

ajasja:
This is excellent, thanks @klimaj ! (Sorry for seeing this so late.) Too bad we can't overload the logical and, or etc, but there are good reasons why not https://peps.python.org/pep-0335/ :)

Issue #29, opened 2024-03-14
Please support gcc version 13.2 and 14.0 in scons - was: gcc 13 compilation error: 'void* __builtin_memmove(void*, const void*, long unsigned int)' forming offset 8 is out of the bounds [0, 8] [-Werror=array-bounds=] ?


smoe:
I do not get what is wrong since 8 is in [0,8]. Any idea what I sould try? I came across something alike
```
#if (__GNUC__ == 13 && __GNUC_MINOR__ == 2)
#pragma GCC diagnostic ignored "-Warray-bounds"
#endif
```
but this may hide something I am not able to see. clang 16 does not have any issue with that particular code.

The following error is reported many times in series until the compilation stops:
```
In file included from /usr/include/c++/13/bits/stl_uninitialized.h:63,
                 from /usr/include/c++/13/memory:69,
                 from src/utility/pointer/std/owning_ptr.hh:20,
                 from src/utility/pointer/owning_ptr.hh:15,
                 from src/protocols/antibody/constraints/ParatopeEpitopeSiteConstraintMover.fwd.hh:18,
                 from src/protocols/antibody/constraints/ParatopeEpitopeSiteConstraintMover.hh:17,
                 from src/protocols/antibody/constraints/ParatopeEpitopeSiteConstraintMover.cc:14:
In static member function 'static _Up* std::__copy_move<_IsMove, true, std::random_access_iterator_tag>::__copy_m(_Tp*, _Tp*, _Up*) [with _Tp = long unsigned int; _Up = long unsigned int; bool _IsMove = false]',
    inlined from '_OI std::__copy_move_a2(_II, _II, _OI) [with bool _IsMove = false; _II = long unsigned int*; _OI = long unsigned int*]' at /usr/include/c++/13/bits/stl_algobase.h:506:30,
    inlined from '_OI std::__copy_move_a1(_II, _II, _OI) [with bool _IsMove = false; _II = long unsigned int*; _OI = long unsigned int*]' at /usr/include/c++/13/bits/stl_algobase.h:533:42,
    inlined from '_OI std::__copy_move_a(_II, _II, _OI) [with bool _IsMove = false; _II = long unsigned int*; _OI = long unsigned int*]' at /usr/include/c++/13/bits/stl_algobase.h:540:31,
[...]

Issue #28, opened 2024-03-14
gcc 13 no longer supports -ffor-scope


smoe:
Hello,

I ran into

```
g++: warning: switch '-ffor-scope' is no longer supported
```
upon
```
scons -j18 cxx=gcc mode=release bin
```
with
```
$ g++ --version | head -n 1
g++ (Debian 13.2.0-13) 13.2.0
```
which I addressed by
```
diff --git a/source/tools/build/basic.settings b/source/tools/build/basic.settings
index 3f0831bad6..d77face5f1 100644
--- a/source/tools/build/basic.settings
+++ b/source/tools/build/basic.settings
@@ -158,7 +158,6 @@ settings = {
                 ],
                 "cxx" : [
                     "std=c++0x",
-                    "ffor-scope",

                     "isystem external/boost_submod/",
                     "isystem external/",
diff --git a/source/update_ResidueType_enum_files.sh b/source/update_ResidueType_enum_files.sh
index bcf505d058..2e8e3ef709 100755
```
but some version-dependency may be preferred?

On a side note, for Debian I had to change the invocations of python in the toplevel scripts executed by scons to python3:
```
--- a/source/update_ResidueType_enum_files.sh
+++ b/source/update_ResidueType_enum_files.sh
@@ -5,7 +5,7 @@
 # (c) For more information, see http://www.rosettacommons.org. Questions about this can be
 # (c) addressed to University of Washington CoMotion, email: license@uw.edu.
 cd "src/core/chemical/residue_properties"
-python update_ResidueType_enum_files.py
+python3 update_ResidueType_enum_files.py
 exitcode=$?
 cd ../../../..
 exit $exitcode
diff --git a/source/update_options.sh b/source/update_options.sh
index aa58d7c83d..33034f907c 100755
--- a/source/update_options.sh
+++ b/source/update_options.sh
@@ -5,7 +5,7 @@
 # (c) For more information, see http://www.rosettacommons.org. Questions about this can be
 # (c) addressed to University of Washington CoMotion, email: license@uw.edu.
 cd "src/basic/options/"
-python options.py
+python3 options.py
 exitcode=$?
 cd ../../..
 exit $exitcode
 ```
Many thanks and regards!
Steffen

smoe:
[...]

PR #36, opened 2024-03-21
Add ability to specify compiler version ranges in scons settings.

roccomoretti:
The basic.settings file has had a limitation for a long while where any version-specific settings are tied to an exact version. That is, settings for gcc 9.0 can't also apply to 10.0 (or even 9.1). This makes things difficult when a potential range of different compilers need to be dealt with in the same fashion, especially in a forward-compatible way. (e.g. the lack of `-ffor-scope` on any GCC newer than 9.0, including ones not yet released)

An adjustment to one of the build files allows us to use a pattern matching specification to add parameter blocks for ranges of compilers (e.g. `cxx_ver:>9.0` for any compiler version greater than 9.0, or `cxx_ver:>=5.2,<7.0` for compiler versions between 5.2 (inclusive) and 7.0 (exclusive).)

Should address issue #29 and #28


lyskov (approved):
LGTM! Rocco, i think it might be a good idea to also schedule all ` gcc-XX.unit` tests to make sure that nothing got broken.

lyskov:
@roccomoretti i also updated Benchmark and added GCC-13 support (unit, debug and release tests) - hope this helps!

PR #39, opened 2024-03-22
libxml2 build: strcasecmp needs strings.h on Debian

smoe:
The exact error (by clang) was that a missing declaration is no longer supported with the latest C++ standards.


roccomoretti (approved):
Looks reasonable.

PR #40, opened 2024-03-22
Fix uninitialized variable in InterfaceScoreCalculator

roccomoretti:
The native_ensemble_best_ variable was uninitialized, which could potentially cause issues. Make sure we initialize it.


MoritzErtelt (approved):
LGTM! Just for my own understanding, previously this worked fine for use within RosettaScripts but outside of that may cause an issue, right?

roccomoretti:
Yes. There's a line which set it in parse_my_tag(), but nothing which initializes it otherwise.

PR #43, opened 2024-03-29
rename torch extras and defines

lyskov:
Current Rosetta `pytorch` and `pytorch_gpu` extras are misnomers since we do link to `libtorch` and not to `pytorch`. This PR address this.


MoritzErtelt (approved):
Looks good to me, that should avoid confusion 👍

roccomoretti:
One concern here would be the documentation and publication issues -- Is there a clear path for people who have used `extras=pytorch` and who are running scripts which explicitly reference `rosetta_scripts.pytorch.linuxgccrelease` going to be able to easily discover why those commands are no longer working?

lyskov:
@MoritzErtelt this is now merged, could you please update the relevant docs? Thanks,

MoritzErtelt:
Sure, I've updated the documentation 👍

PR #49, opened 2024-04-16
Test and fix for mp_f19_decoy_discrimination failure

roccomoretti:
The mp_f19_decoy_discrimination scientific test has a hard failure. It's crashing with the following issue in the minimization routine of relax. 
```
core.scoring.hbonds.hbonds_geom: [ ERROR ] NAN occurred in H-bonding calculations!
core.scoring.hbonds.hbonds_geom: [ ERROR ] Dxyz                    -NAN                    -NAN                    -NAN  Hxyz                    -NAN                    -NAN                    -NAN
```

This PR pulls out the offending run/structure as an integration test. 

Playing around with it, I figured out two things
1. The `-constrain_relax_to_start_coords true` flag is needed to show symptoms
2. It was specifically the derivative calculations which were causing NaN issues.

Inserting a number of `std::isnan()` checks and trying to trace back the source indicates the problem was `MembraneGeometry::g_radius_gradient_dz()`, specifically where it did a `(p.x()-xo)/std::abs(p.x()-xo))`, which results in a NaN if p.x() equals xo.  We can rearrange this slightly, not only to avoid the NaN, but also to reduce the number of calculations needed. While I was at it, additional efficiency changes in MembraneGeometry were also made.


roccomoretti:
> Was the 90 test done once?

Yes. https://b3.graylab.jhu.edu/revision/rosetta:commits/20383

I noticed I didn't actually run the mp_f19_decoy_discrimination test on the server, so that's now queued as well: https://b3.graylab.jhu.edu/test/833171

roccomoretti:
@Rituparnasamanta Any further thoughts, or additional corrections/tests I should perform?

lyskov (approved):
LGTM!

PR #51, opened 2024-04-19
Update benchmark scripts for Python3.12.

roccomoretti:
Python 3.12 gets rid of the `imp` module. Moreover, it gets rid of `imp.load_source()` altogether. We thus need to change around how we do the magic 'import from parent directory' statement.

There are also a few other things which were popping up warnings when I ran things locally.

### DIFF
--- a/tests/benchmark/tests/PyRosetta.py
+++ b/tests/benchmark/tests/PyRosetta.py
@@ -12,9 +12,6 @@
 ## @brief  PyRosetta binding self tests
 ## @author Sergey Lyskov
 
-# A bit of Python magic here, what we trying to say is this: from ../__init__ import *, but init path is calculated relatively to this location
-exec(open('/'.join(__file__.split('/')[:-1]) +  '/__init__.py').read(), globals())
-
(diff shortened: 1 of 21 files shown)


tests/benchmark/tests/PyRosetta.py
```diff
@@ -12,6 +12,9 @@
 ## @brief  PyRosetta binding self tests
 ## @author Sergey Lyskov
 
+# A bit of Python magic here, what we trying to say is this: from ../__init__ import *, but init path is calculated relatively to this location
+exec(open('/'.join(__file__.split('/')[:-1]) +  '/__init__.py').read(), globals())
+
```
lyskov, line 17: @roccomoretti i think this will be the exact replacement if we want to avoid using direct `exec`:

```Python
from importlib.machinery import SourceFileLoader
SourceFileLoader(__name__, '/'.join(__file__.split('/')[:-1]) +  '/__init__.py').load_module()
```

lyskov (approved):
LGTM, - thank you for fixing this @roccomoretti !

PR #52, opened 2024-04-23
Fix windows build.

roccomoretti:
Make use of POSIX header platform dependent.
I believe that io.h is the Windows equivalent for getting the read() function.


lyskov (approved):
LGTM!

PR #54, opened 2024-04-23
Update SCons for compilation with Python 3.12

roccomoretti:
Python 3.12 removed the `imp` module, which means our current SCons 3.0.4 doesn't work with it.

Upgrading to SCons 4 will fix this. However, this removes support for building with Python 2.7 -- I think at this point this is acceptable, as anyone compiling Rosetta will likely have a Python3 version installed for RFDiffusion, etc. anyway. Additionally, the error message printed is relatively user friendly:

```
scons: *** SCons version 4.2.0 does not run under Python version 2.7.
Python >= 3.5 is required.
```

SCons 4.2.0 was selected as it's the last version which supports Python 3.5 -- I don't know how many people that will actually help, but we don't need to be bleeding edge.

Some additional files were changed to address warning messages with a Python 3.12 build. 

(Note that the current external/scons-local directory was deleted and then re-created from the official download. No manual changes were made, save for deleting some unnecessary documentation files.)


lyskov:
> (Note that the current external/scons-local directory was deleted and then re-created from the official download. No manual changes were made, save for deleting some unnecessary documentation files.)

@roccomoretti maybe we should put into README file for our own record? Like `source/external` perhaps?

lyskov (approved):
LGTM, - thank you for taking care of this Rocco!

lyskov (approved):
LGTM!

PR #55, opened 2024-04-23
Clean up integration tests for running on Python 3.12

roccomoretti:
If we're updating compilation to only use Python3 (#54), updating the integration tests to remove Python2 dependence also makes sense.

It looks like we've mostly fixed any usage of Python2 already, but I can clean up the ability to use python2 in the framework.

There's also some additional fixes to be compatible with Python 3.12 in particular, as well as some additional diagonstics that can be printed which might be helpful for running on a fresh copy of Python (e.g. a simple Conda environment).

I also cleaned up the test commands such that they all use the same python which is used to launch the integration.py script. (Now that we don't need to have multiple pythons.)


lyskov (approved):
LGTM!

PR #56, opened 2024-04-25
removing __future__ statements

NooraAz:
No more relevant in python3.


roccomoretti:
They certainly don't harm any Python3 run, and do help if someone were to run under Python2 for some reason. I don't see the need to have a PR just to remove them.

PR #57, opened 2024-04-25
fixing util file names

NooraAz:
The argument passed to --class_name will be used as the prefix of util files.

### DIFF
--- a/source/code_templates/generate_templates.py
+++ b/source/code_templates/generate_templates.py
@@ -17,7 +17,7 @@ from __future__ import print_function
 from __future__ import absolute_import
 import os
 work_dir = os.getcwd()
-
+# change the directory to this file's directory, if it is run from elsewhere.
 if os.path.dirname(__file__): os.chdir(os.path.dirname(__file__))
 from argparse import ArgumentParser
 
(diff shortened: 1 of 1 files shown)


roccomoretti:
This would change the current default behavior for the util files, right? (What happens if you run the script for util without --class_name  specified?) 

I'm wondering if there's a way you can keep the current behavior with the command lines which currently work, but add in support for the additional behavior if you add in additional options.

vmullig:
> I'm wondering if there's a way you can keep the current behavior with the command lines which currently work, but add in support for the additional behavior if you add in additional options.

Since the old behaviour was counter-intuitive (the class name attribute was used to set the file name in all cases _except_ making utility files, and this resulted in `util.hh` and `util.cc` being overwritten), I think replacing this with the new behaviour is fine -- especially since this isn't something that came up frequently, and a developer using this functionality is bound to run the `generate_templates.py` script with `--help` first.  Noora has added a clear description to the `--help` output about what the new behaviour is.  This looks pretty good to me.

vmullig (approved):
Looks good to me!

lyskov:
Looks like code_template_tests_src is still broken, - @NooraAz could you please look this up? Thanks,

NooraAz:
> Looks like code_template_tests_src is still broken, - @NooraAz could you please look this up? Thanks,

@lyskov as Rocco mentioned, it is because in the test `class_name` is not passed in utils, while I set it to be required for all file types (including utils). So I think the test should change, but not sure how I can do that. (unless you don't think this is the way to go).

lyskov:
Looping in @jadolfbr who originally wrote this script, - Jared what do you think best path forward here? Thanks,

ajasja:
Sorry for not catching that, I think it looked like all tests passed.

PR #58, opened 2024-04-29
Fix pre-compile script calls on Windows build

roccomoretti:
The windows build test on the test server doesn't seem to be updating options appropriately. 

It looks like the issue is that the *.sh files aren't being properly called -- adding "bash" to the front of them allows them to run properly:

Before:
```
Running versioning script ... Done. (0.4 seconds)
```

After:
```
Running versioning script ... Done. (0.4 seconds)
file ./options.dox being updated
file ./full-options-list.md being updated
Number of option files updated: 2
Total 4680 options.
Updating submodules needed for compilation.
Finished updating ResidueProperty code
-- no changes needed
Finished updating VariantType code
-- no changes needed
```


lyskov (approved):
Thank you for fixing this @roccomoretti ! Its a bit odd that current (not correct) behavior has not resulted in build failure, - any thoughts on that?

roccomoretti:
No. It's somewhat surprising.

PR #59, opened 2024-05-09
Updated Ligand Motifs Code

toadlover:
The code I am contributing here are updates to improve the functionality of the ligand motif code. This includes improvements to the following protocols:

motifs/LigandDiscoverySearch - Created a method for scoring placed ligands based on how much of a user-defined binding pocket is filled, improved use of score functions, collection and analysis of motifs collected off of placed ligands to confirm contacts against residues of interest

motifs/IdentifyLigandMotifs - Moved usage of process_for_motifs from private to public so that the function to collect motifs can be used on poses in runtime (and not exclusively from imported files). The process_for_motifs function is now used in LigandDiscoverySearch.

Added new options in options_rosetta.py under motifs for use in LigandDiscoverySearch.

Added public app remove_duplicate_motifs.cc, which removed motifs that are considered duplicates based on similar distance and angle of atoms involved. This helps improve runtimes of protocols that use lists of motifs by not considering the use of motifs that are effectively very similar/same. I have added an integration test as well.


source/src/protocols/motifs/IdentifyLigandMotifs.hh
```diff
@@ -114,7 +117,8 @@ private:
 		std::string & pdb_name,
 		protocols::motifs::MotifLibrary & motifs,
 		core::scoring::ScoreFunctionOP scorefxn,
-		utility::vector1< utility::vector1< core::Size > > & motif_indices_list
+		utility::vector1< utility::vector1< core::Size > > & motif_indices_list,
+		utility::vector1< core::Size > & prot_pos_that_made_motifs
```
roccomoretti, line 121: const correctness helps with documentation. If you're not going to change a value that's passed by reference, it's best to pass it by `const &` to annotate that fact. If you're consistent in doing so, it makes the fact that you _are_ modifying prot_pos_that_made_motifs stick out more.

roccomoretti (commented):
It looks like you're still changing this PR, so this is an in-progress comments only.

source/src/protocols/motifs/LigandDiscoverySearch.hh
```diff
@@ -182,6 +223,30 @@ private:
...
+
+	// @brief a boolean value that controls the verbosity of text output while using the object
+	//default to be false, but can be made true by using the motifs::verbose flag; can also use set_verbose()
+	int verbose_;
+
+	// @brief a vector to act as a cipher for the space fill matrix with naming atoms with a unique name
+	//due to there being a limit in an atom name only having up to 4 characters and ther ebeign potential for a full representation matrix to have 1e5-1e6 atoms, using a base62 system for unique names
+	//base 62 allows over 14 million unique names
+	//represented as alphanumeric characters 0-9,a-z,A-Z
+	utility::vector1<std::string> base_62_cipher_{"0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k", "l", "m", "n",
```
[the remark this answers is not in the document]
toadlover: It looks like that conversion could work in theory (and I like that it is in base 64 instead of 62). However, I do not like that "/" is in the character set, and fear it would not work for what I need. I am using this function for unique atom naming conventions that would be written into PDB files, and my understanding is that a slash could hurt parsing. If it was a more friendly character (- or _), that would be more welcome. Would a compromise be to make another function in  src/utility/Binary_Util.hh with a slightly modified character set that does not use the backslash? Perhaps even an option to use a custom 64 character set?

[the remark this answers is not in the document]
roccomoretti: Oh, if the provided character set doesn't work, then there isn't a need to combine the two.

You could potentially expand the current function to be flexible with respect to the character set, but that isn't needed if you don't think it's worth it.

I would add a comment as to why you're rolling your own, though.

toadlover:
Hi, apologies on the long delay. I think my code is ready for another look, correcting the recommended changes. Looking at the benchmark server, I don't think any of the failed tests are problematic (the worst is likely just because of the tiny fix in the trig function with accumulated error.

source/src/protocols/motifs/LigandDiscoverySearch.cc
```diff
@@ -292,694 +366,1790 @@ utility::vector1<core::conformation::ResidueOP> LigandDiscoverySearch::get_all_r
...
+// @brief create a constraint set on the 3 ligand motif atoms (the last residue in working_pose_) to reduce their movement before applying a highresdock
+void LigandDiscoverySearch::add_constraints_to_working_pose(const core::Size trip_atom_1, const core::Size trip_atom_2, const core::Size trip_atom_3, const core::Size working_position, const core::conformation::ResidueOP ligresOP)
+{
+	//create constraints for ligand wiggling to add to the pose
+	constraints::ConstraintSetOP sc_cst_set( new constraints::ConstraintSet() );
 
-		// This is to make an atomtypeset to get atomtype integers
-		core::chemical::AtomTypeSetCOP atset = core::chemical::ChemicalManager::get_instance()->atom_type_set( FA_STANDARD );
+	core::scoring::func::FuncOP fx1( new core::scoring::func::HarmonicFunc( 0.0, 1.0 ) );
+	sc_cst_set->add_constraint( core::scoring::constraints::ConstraintCOP( utility::pointer::make_shared< core::scoring::constraints::CoordinateConstraint >( core::id::AtomID( trip_atom_1, working_pose_->size() ), core::id::AtomID( working_pose_->residue( working_position ).atom_index( "CA" ), 1 ), ligresOP->xyz( trip_atom_1 ), fx1 ) ) );
```
roccomoretti, line 547: `utility::pointer::make_shared<core::scoring::constraints::CoordinateConstraint>` will create a new OP of the type `core::scoring::constraints::CoordinateConstraintOP`, you don't have to manually cast it to `core::scoring::constraints::ConstraintCOP` -- it should be automatically convertable to the base-class OP, and the OP -> COP transition should also happen automatically.

source/src/protocols/motifs/LigandDiscoverySearch.cc
```diff
@@ -292,694 +366,1790 @@ utility::vector1<core::conformation::ResidueOP> LigandDiscoverySearch::get_all_r
...
+void LigandDiscoverySearch::add_ligand_to_pose_residuetypeset(const core::chemical::MutableResidueTypeOP lig_mrt)
+{
+	//code to add type set of imported ligand into pose
+	core::chemical::PoseResidueTypeSetOP rts( working_pose_->conformation().modifiable_residue_type_set_for_conf( core::chemical::FULL_ATOM_t ) );
 
-						utility::vector1< core::Size > atom_j_vector;
-						atom_j_vector.push_back( atom_i_connects[atom_j] );
-						atom_j_vector.push_back( atset->atom_type_index( ligresOP->atom_type(atom_i_connects[atom_j]).atom_type_name() ) );
+	//make a ResidueTypeSetCOP that will be pulled from the PoseResidueTypeSetOP
+	core::chemical::ResidueTypeSetCOP def_rts(rts->default_rts());
```
roccomoretti, line 576: I wouldn't bother to grab the default RTS -- you can check the PoseRTS for the presence of the ligand directly. (Which is likely more what you want, as if the PoseRTS has the ligand but the underlying typeset doesn't, then you probably still want to skip adding it.)

source/src/protocols/motifs/LigandDiscoverySearch.cc
```diff
@@ -292,694 +366,1790 @@ utility::vector1<core::conformation::ResidueOP> LigandDiscoverySearch::get_all_r
...
+	}
 
-						ligand_atom_trios.push_back(cur_motif_indices);
-					}
-				}
-			}
-		}
+	//reset the residue type sets for the working_pose_ and the original_pose_ so that they are updated to have the new ligand
+	working_pose_->conformation().reset_residue_type_set_for_conf(rts);
+	original_pose_.conformation().reset_residue_type_set_for_conf(rts);
```
roccomoretti, line 589: You likely only need to reset things if you've actually changed it.

source/src/protocols/motifs/LigandDiscoverySearch.cc
```diff
@@ -292,694 +366,1790 @@ utility::vector1<core::conformation::ResidueOP> LigandDiscoverySearch::get_all_r
...
+					run_HighResDock_on_working_pose(my_HighResDocker);
+
+					//use fa atr/rep or whole function based on highresdock_with_whole_score_fxn flag to get ddg before highresdock
+					if ( option[ OptionKeys::motifs::highresdock_with_whole_score_fxn ] ) {
+						//whole
+						delta_score = get_pose_ddg(whole_score_fxn_, working_pose_);
+					} else {
+						//atrrep
+						delta_score = get_pose_ddg(fa_atr_rep_fxn_, working_pose_);
+					}
```
roccomoretti, line 1135: Might simplify things if you have something like a `working_score_fxn_` ScoreFunctionOP member variable, and set that at initialization time to either `whole_score_fxn_` or `fa_atr_rep_fxn_` based on the option. Then you can just use `working_score_fxn_` directly, without having to re-consult the option in multiple locations.

source/src/protocols/motifs/LigandDiscoverySearch.cc
```diff
@@ -292,694 +366,1790 @@ utility::vector1<core::conformation::ResidueOP> LigandDiscoverySearch::get_all_r
...
+	return 1;
+}
+
+// @brief A function to reset the working_pose_ back to its original state, as found in original_pose
+//to call this function the least amount of times (and minimize wasted time/resources), this function is only to be called right before a placement is culled before a continue statement after working_pose_ had a ligand added to it
+void LigandDiscoverySearch::reset_working_pose()
+{
+	core::pose::PoseOP original_poseop(original_pose_.clone());
+	//assign the original_poseop to working_pose_
+	working_pose_ = original_poseop;
```
roccomoretti, line 1439: `working_pose_ = original_pose_.clone();` works.

source/src/protocols/motifs/LigandDiscoverySearch.cc
```diff
@@ -292,694 +366,1790 @@ utility::vector1<core::conformation::ResidueOP> LigandDiscoverySearch::get_all_r
...
+	return motif_holder;
+}
+
+//create protein_representation_matrix_
+//uses working_pose to make the matrix
+void LigandDiscoverySearch::create_protein_representation_matrix(core::Size & x_shift, core::Size & y_shift, core::Size & z_shift, int & x_bound_int, int & y_bound_int, int & z_bound_int)
+{
+
+	//create tracer to identify points of the run
+	static basic::Tracer ms_tr( "LigandDiscoverySearch.create_protein_representation_matrix", basic::t_info );
```
roccomoretti, line 1477: Having a tracer per function might be a bit overkill, unless you think you really need that granularity in control. (Usually a class level assignment of error/warning/info/debug/trace is sufficient -- if you need debug level output for one function in the class, getting debug-level output for all the other functions usually doesn't hurt anything, and may help.)

source/src/protocols/motifs/LigandDiscoverySearch.cc
```diff
@@ -292,694 +366,1790 @@ utility::vector1<core::conformation::ResidueOP> LigandDiscoverySearch::get_all_r
...
+
+	//determine largest and smallest x,y,z  values to determine dimensions of matrix
+	for ( core::Size res_num = 1; res_num <= working_pose_->size(); ++res_num ) {
+		for ( core::Size atom_num = 1; atom_num <= working_pose_->residue(res_num).natoms(); ++atom_num ) {
+			//get the x,y,z data of the atom
+			numeric::xyzVector<int> atom_xyz = working_pose_->residue(res_num).xyz(atom_num);
+			//convert the coordinates to integers
+			atom_xyz.x() = static_cast<int>(atom_xyz.x());
+			atom_xyz.y() = static_cast<int>(atom_xyz.y());
+			atom_xyz.z() = static_cast<int>(atom_xyz.z());
```
roccomoretti, line 1503: You've already cast to integers in the line above. atom_xyz can't hold anything except integers, so all the values need to be converted during assignment.

roccomoretti: BTW, hopefully you're aware of the rules that C++ uses to truncate Real numbers to integers, and you're okay with the behavior on both the positive and negative side of zero.

source/src/protocols/motifs/LigandDiscoverySearch.cc
```diff
@@ -292,694 +366,1790 @@ utility::vector1<core::conformation::ResidueOP> LigandDiscoverySearch::get_all_r
...
+	//derive constant values based  on the most negative values in each dimension, and then add that constant to all coordinates
+
+	int smallest_x = 1;
+	int smallest_y = 1;
+	int smallest_z = 1;
+
+	int largest_x = 1;
+	int largest_y = 1;
+	int largest_z = 1;
+
```
roccomoretti, line 1490: Just a note that PDBs aren't guaranteed to be centered around 0. You can theoretically have a protein structure which has z coordinates which range from -501 to -554, or x coordinates which range from 356 to 467. Starting with each at 1 might not be what you want.

source/src/protocols/motifs/LigandDiscoverySearch.cc
```diff
@@ -292,694 +366,1790 @@ utility::vector1<core::conformation::ResidueOP> LigandDiscoverySearch::get_all_r
...
+			atom_coordinates.push_back(atom_xyz);
+
+		}
+	}
+
+	//take negative values of the smallest values and then add 1 to derive the constants
+	x_shift = (smallest_x * -1) + 1;
+	y_shift = (smallest_y * -1) + 1;
+	z_shift = (smallest_z * -1) + 1;
```
roccomoretti, line 1535: Again, `smallest_x` could end up either positive or negative. Check to make sure this is doing what you want in both cases.

source/src/protocols/motifs/LigandDiscoverySearch.cc
```diff
@@ -292,694 +366,1790 @@ utility::vector1<core::conformation::ResidueOP> LigandDiscoverySearch::get_all_r
...
+		}
+		//push a 2D  matrix into the 3D matrix
+		protein_representation_matrix_.push_back(sub_matrix);
+	}
+
+	//seed the matrix with approximate coordinates of each atom
+	//approximated by always rounding down via int casting
+	for ( core::Size xyzVec = 1; xyzVec <= atom_coordinates.size(); ++xyzVec ) {
+		protein_representation_matrix_[static_cast<int>(atom_coordinates[xyzVec].x())][static_cast<int>(atom_coordinates[xyzVec].y())][static_cast<int>(atom_coordinates[xyzVec].z())] = true;
+	}
```
roccomoretti, line 1578: You don't do any shift on these --- the coordinate indexes could end up negative.

source/src/protocols/motifs/LigandDiscoverySearch.cc
```diff
@@ -292,694 +366,1790 @@ utility::vector1<core::conformation::ResidueOP> LigandDiscoverySearch::get_all_r
...
+//void LigandDiscoverySearch::create_protein_representation_matrix_space_fill(core::Size & x_shift, core::Size & y_shift, core::Size & z_shift, core::Size & x_bound_int, core::Size & y_bound_int, core::Size & z_bound_int, int & resolution_increase_factor,
+//core::Size & sub_x_min, core::Size & sub_x_max, core::Size & sub_y_min, core::Size & sub_y_max, core::Size & sub_z_min, core::Size & sub_z_max, core::Real & occupied_ratio, core::Real & sub_occupied_ratio)
+//condensing arguments in function to use vectors to hold xyz trios
+void LigandDiscoverySearch::create_protein_representation_matrix_space_fill(utility::vector1<core::Size> & xyz_shift, utility::vector1<core::Size> & xyz_bound, int & resolution_increase_factor,
+	utility::vector1<core::Size> & sub_xyz_min, utility::vector1<core::Size> & sub_xyz_max, utility::vector1<core::Real> & occupied_ratios, utility::vector1<core::Size> & matrix_data_counts, core::Size working_position)
+{
+	//values that can be seeded into the matrix with their meaning. Only 0-3 can be seeded using this function, and then space_fill_analysis can seed 4+
+	//note, 0 and 2 are only even numbers that are currently used for now
+	/*
```
roccomoretti, line 1591: This function is copying a bunch of functionality from the previous one. I'd highly recommend abstracting out the commonalities into a common protein grid class.

Better yet, take a look at existing functionality like `core/grid/CartGrid.hh` and see if it would work for your purpose.

roccomoretti (commented):
Some opportunities to clean things up.

Also, to get a better read on the test server, you probably want to merge with a recent `main`, such that you're properly comparing to a recent version of main.

source/src/protocols/motifs/LigandDiscoverySearch.cc
toadlover: Yeah, it probably is a good idea to go ahead and just extract this logic into its own class, and then others can use it for purposes outside of my motif discovery code. I have looked a little into the CardGrid code, and will look more into it before making a final decision. I am leaning towards making these protein matrices their own class, since these are specifically for wrapping around and simplifying a protein system. Making the protein grid an inherited class off of the CardGrid may be prudent, and I'll look into seeing how compatible the base class is. If I do create a new class, inherited or not, is it reasonable to put the code in the core/grid location, or would elsewhere be better?

source/src/protocols/motifs/LigandDiscoverySearch.cc
toadlover: The shift occurs a few lines further up in the for loop where I iterate over atom_coordinates and add the corresponding shift to the coordinate. I can organize this better so that the shift operation is more apparent.

source/src/protocols/motifs/LigandDiscoverySearch.cc
toadlover: I'm adapting your prior points, as I had not accounted as well as I should have for operating on either side of zero, and to make sure that the smallest and largest values are more safely seeded. I am now taking the floor, instead of truncating, so that every value gets rounded down, regarded of being positive or negative. With that, I think this shift value block is correct as-is. I'm elaborating on it further in comments to better explain its purpose. The purpose of these values are to apply a constant shift in the XYZ coordinates of each atom, where the absolute smallest value in any of the dimensions are set to 1. By adding the negative of the smallest value to itself, that results in 0, and then 1 is added to shift to have the minimum be 1 so that it is the minimum of the vector1.

source/src/protocols/protein_grid/ProteinGrid.cc
```diff
@@ -0,0 +1,1493 @@
...
+	using_lj_radii_ = false;
+	matrix_fullness_ = 0;
+	fullness_ratio_ = 0;
+	sub_matrix_fullness_ = 0;
+	sub_fullness_ratio_ = 0;
+
+	//seed variables that are relevant for how to output the matrix to a pdb for visualization
+	//not planning to have devoted constructors to define these variables
+	print_whole_matrix_ = basic::options::option[ basic::options::OptionKeys::protein_grid::output_whole_matrix ]();
+	print_empty_space_ = basic::options::option[ basic::options::OptionKeys::protein_grid::output_empty_space ]();
```
roccomoretti, line 238: Consulting the option system isn't possible with in-class initialization, so setting these with a function make sense, though convention is to be explicit in the function name (i.e. `initialize_from_options()`)

source/src/protocols/protein_grid/ProteinGrid.cc
```diff
@@ -0,0 +1,1493 @@
...
+	using_sub_area_ = true;
+
+	//wrap matrix around pose
+	wrap_matrix_around_pose();
+}
+
+//copy constructor
+ProteinGrid::ProteinGrid( ProteinGrid const & other )
+{
+	initialize();
```
roccomoretti, line 181: If the clone function is just going to overwrite the initialized variables, then this call is pointless. And if the clone function is simply a straight member-by-member copy, it's much better to use the compiler-provided default one,  (`ProteinGrid::ProteinGrid( ProteinGrid const & other ) = default;` which has the benefit of being automatically updated whenever someone adds a new member variable.

Similar arguments apply for `operator=()` -- prefer the compiler default unless you have very special use case where it doesn't work.

source/src/protocols/protein_grid/ProteinGrid.cc
```diff
@@ -0,0 +1,1493 @@
...
+		dummylig_mrt->delete_atom(dummylig_mrt->all_atoms()[1]);
+	}
+
+
+	ms_tr.Trace << "Number of atoms in dummy mrt is now: " <<  dummylig_mrt->natoms() << std::endl;
+
+	//delete all chis in the dummy ligand
+	//probably easiest to use delete_terminal_chi
+	while ( dummylig_mrt->nchi() > 0 )
+			{
```
roccomoretti, line 851: I'm surprised the location of the curly brace isn't causing the beautifier to complain.

source/src/protocols/protein_grid/ProteinGrid.cc
```diff
@@ -0,0 +1,1493 @@
...
+	return base_62_number;
+}
+
+// @brief default constructor
+//will need to use class functions to seed values for input pose and other potential input data
+//should only use parameterized or copy constructor
+ProteinGrid::ProteinGrid()
+{
+
+}
```
roccomoretti, line 1210: If you're not using the default constructor for anything, it's best just to omit it. (As you'll get a compiler error if you ever accidentally try to invoke it.)

source/src/protocols/protein_grid/ProteinGrid.hh
```diff
@@ -0,0 +1,248 @@
...
+
+#include <ObjexxFCL/string.functions.hh>
+
+#include <basic/Tracer.hh>
+
+#include <algorithm>
+#include <sstream>
+
+namespace protocols {
+namespace protein_grid {
```
roccomoretti, line 41: Given that the ProteinGrid seems specialized for the LigandDiscoverySearch use case, I'm not sure that you need to make a new protocols directory for it. (That is, it doesn't look like someone would use ProteinGrid without also using the motifs/ code.)

roccomoretti (commented):
Pretty good -- just some minor suggestions for cleanup, mostly related to the ProteinGrid constructors.

roccomoretti:
Oh, comments on test results:

Looks like you're getting hard failures in the real_motif_analysis and remove_duplicate_motifs tests in the release_debug runs (https://b3.graylab.jhu.edu/test/855709) -- these aren't just "a reference run isn't available" failures, but failures that indicate things are crashing/failing even before the comparison. I'm not sure why they are failing for release_debug but not the regular release runs, but it may be due to the extra checks in debug mode. (Or it could be a Mac/Linux difference.)

Delete the `//#include <protocols/motifs/FindMotifPositionsOnLigandHelper.cc>` line in `src/apps/public/motif_ligand_discovery/remove_duplicate_motifs.cc` (even commented out, it's tripping up the include_cc_check)

source/src/protocols/protein_grid/ProteinGrid.hh
toadlover: I would disagree, and say that this does have potential usage beyond the LigandDiscoverySearch, and beyond general usage of motifs and drug docking. I do agree that the class is currently more specialized for my LDS code, but could be expanded upon further for additional functionality with other protocols that may need it. The clash check is more catered towards the LDS (and checking for ligand-backbone clashes), but could maybe be used if someone else wants to check if a ligand would clash with a protein in a more general case. The space fill analysis I think is what is more broadly applicable, as it can be used to quantify how occupied a region of a protein is by protein and/or ligand (including the additional code for visualization). I'm not 100% opposed to moving it into the motifs code (and can move it in if you still feel that it should be), but do feel that it has a broader usage case that could warrant keeping it in a different space.

source/src/protocols/protein_grid/ProteinGrid.cc
toadlover: The beautifier seems to keep trying to push the bracket to this location. Is it best to just leave it where the beautifier likes?

source/src/protocols/protein_grid/ProteinGrid.cc
```diff
@@ -175,63 +171,39 @@ ProteinGrid::ProteinGrid(core::pose::PoseOP in_pose, core::Real resolution, nume
...
+	this->true_sub_region_dimensions_ = copy.true_sub_region_dimensions_;
+	this->adjusted_sub_region_dimensions_ = copy.adjusted_sub_region_dimensions_;
+	this->sub_region_max_ = copy.sub_region_max_;
+	this->sub_region_min_ = copy.sub_region_min_;
+	this->matrix_fullness_ = copy.matrix_fullness_;
+	this->fullness_ratio_ = copy.fullness_ratio_;
+	this->sub_matrix_fullness_ = copy.sub_matrix_fullness_;
+	this->sub_fullness_ratio_ = copy.sub_fullness_ratio_;
+	this->print_whole_matrix_ = copy.print_whole_matrix_;
+	this->print_empty_space_ = copy.print_empty_space_;
```
roccomoretti, line 200: If this is used somewhere, you just changed the semantics.

If it's not used anywhere, I'd recommend just deleting this function. (The functionality is better handled by the copy constructor/assignment operator.)

roccomoretti (approved):
Looks like there's a bit of instability in the intergration.mpi tests, but there's nothing particular to MPI here, and the non-MPI tests look fine. -- Should keep an eye on it for the post-merge results, but I don't think there's a need to hold off on the merge because of it.

toadlover:
Merging in significant update to LigandDiscoverySearch code, including helper class: ProteinGrid, which is now used to both detect backbone clashing and quantify the fullness of a binding pocket.

toadlover:
> Looks like there's a bit of instability in the intergration.mpi tests, but there's nothing particular to MPI here, and the non-MPI tests look fine. -- Should keep an eye on it for the post-merge results, but I don't think there's a need to hold off on the merge because of it.

Sounds good! It looks like merging is still blocked on my end.

PR #63, opened 2024-06-04
Add -Wno-error for external libraries.

roccomoretti:
We shouldn't be doing warnings-as-errors for external libraries.

This change should make things more explicit.


lyskov (approved):
👍

PR #64, opened 2024-06-05
Fix issue with JobGenealogist and serialization.

roccomoretti:
An attempt at modernizing the serialization method for protocols::jd3::compare_job_nodes accidentally made things be called in an infinite loop. This is breaking the GCC 12 serialization build.

The fix is easy -- there's now no data to serialize, so we can simply replace the functions with empty ones.


lyskov (approved):
LGTM, - thank you for taking care of this @roccomoretti !

PR #95, opened 2024-07-17
Rmsd in smallmol docking

Hackefleisch:
This PR does two things:

1. It fixes the RMSD accept/reject option in the TransformMover. The old version did not square the distances and was not automorphic. The new version simply uses the Rosetta automorphic rmsd function. Additionally, there is now an informative print to tell a user how many moves were rejected due to rmsd criteria.
2. The RMSD criteria is copied 1 to 1 to the HighResMover and follows the same functionality.

Both changes are useful to limit docking exploration during execution towards the starting position and enforces local refinements.

### DIFF
--- a/source/src/protocols/ligand_docking/HighResDocker.cc
+++ b/source/src/protocols/ligand_docking/HighResDocker.cc
@@ -44,6 +44,7 @@
 #include <core/pack/rotamer_set/UnboundRotamersOperation.hh>
 
 #include <core/scoring/ScoreFunction.hh>
+#include <core/scoring/rms_util.hh>
 
 // Package Headers
 #include <core/pose/Pose.hh>
@@ -127,7 +128,9 @@ HighResDocker::HighResDocker(HighResDocker const & that):
 	num_cycles_(that.num_cycles_),
 	repack_every_Nth_(that.repack_every_Nth_),
 	score_fxn_(that.score_fxn_),
-	movemap_builder_(that.movemap_builder_)
+	movemap_builder_(that.movemap_builder_),
+    rmsd_max_(that.rmsd_max_),
+    check_rmsd_(that.check_rmsd_)
 {
 	initialize_from_options();
 }
@@ -185,6 +188,12 @@ HighResDocker::parse_my_tag(
 	if ( tag->hasOption("resfile") ) {
 		resfile_= tag->getOption<std::string>("resfile");
 	}
+
+    /// RMSD ///
+    if ( tag->hasOption("rmsd") ) {
+        check_rmsd_ = true;
+        rmsd_max_ = tag->getOption<core::Real>("rmsd");
+    }
 }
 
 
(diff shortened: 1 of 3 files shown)


source/src/protocols/ligand_docking/HighResDocker.cc
```diff
@@ -282,11 +291,26 @@ HighResDocker::apply(core::pose::Pose & pose) {
...
+	utility::vector1<protocols::moves::MoverOP> rigid_body_movers = create_rigid_body_movers(pose);
+
+    core::Size ligand_index = 0;
+    for ( core::Size i(1), imax(pose.total_residue()); i<=imax; ++i ) {
+        core::conformation::Residue const & residue( pose.residue(i) );
+        if ( residue.is_ligand() ) {
+            ligand_index = i;
+        }
+    }
+    core::conformation::Residue original_ligand(pose.residue(ligand_index));
```
roccomoretti, line 303: Prefer using a ResidueOP and `.clone()` to make a copy. I don't think it makes a difference here, but it does in other contexts where there's multiple derived types, it plays better with subtypes (the approach here would lose subtype info.)

source/src/protocols/ligand_docking/Transform.cc
```diff
@@ -612,12 +619,9 @@ bool Transform::check_rmsd(core::conformation::UltraLightResidue const & start,
 {
 	debug_assert(start.natoms() == current.natoms());
 
-	core::Real total_distance =0.0;
-	for ( core::Size atomno = 1; atomno <= start.natoms(); ++atomno ) {
-		total_distance += start[atomno].distance(current[atomno]);
-	}
+	core::Real rmsd = core::scoring::automorphic_rmsd( *start.residue(), *current.residue(), /*superimpose=*/false );
```
roccomoretti, line 622: Part of the purpose of UltraLightResidue is to not have to update the coordinates of the underlying conformation::Residue and take the overhead of that. As such, the `residue()` coordinates are set at initialization and not updated -- your change here isn't doing what you want it to (it's using stale coordinates).

source/src/protocols/ligand_docking/HighResDocker.cc
```diff
@@ -282,11 +291,26 @@ HighResDocker::apply(core::pose::Pose & pose) {
...
 	// I believe that this may have been related to adding constraints incorrectly at other places in my code.
 	// Rigid body exploration
-	utility::vector1<protocols::moves::MoverOP> rigid_body_movers= create_rigid_body_movers(pose);
+	utility::vector1<protocols::moves::MoverOP> rigid_body_movers = create_rigid_body_movers(pose);
+
+    core::Size ligand_index = 0;
+    for ( core::Size i(1), imax(pose.total_residue()); i<=imax; ++i ) {
+        core::conformation::Residue const & residue( pose.residue(i) );
+        if ( residue.is_ligand() ) {
+            ligand_index = i;
```
roccomoretti, line 300: There can theoretically be multiple ligands in the pose, only one of which (not necessarily the last) is the one which is being modified. Unfortunately, things are indirect enough that I don't think there's a good way of getting just the primary ligand from the info provided, so there likely isn't a need to change things. Just make sure to document the issue thoroughly. -- Though I might suggest breaking this out into a private function, which might make updating things later easier.

roccomoretti (commented):
You should add the test labels when this is ready to go.

I haven't bothered, as it looks like it will fail the beautification tests. (Use tabs instead of spaces to indent Rosetta C++ code.)

PR #107, opened 2024-08-02
Fix compilation issue with SmartFixbbSimAnnealer.cc

roccomoretti:
Discussion #106 indicated that we may be missing an include for std::array in the USE_TENSORFLOW case.


lyskov (approved):
LGTM!

PR #115, opened 2024-08-12
Fix issue related to PackerTask usage in homodimer_design.

roccomoretti:
With a -pack_min_runs of 2 or greater, there's a chance that the sequence of the structure changes. This is a problem for a fixed PackerTask run, as `task_is_valid( pose )` fails if the base residue types don't match up properly. This is an issue found in Discussion #112 

The fix is simple -- just use the TaskFactory which we already have lying around instead of manually generating the packer task. (The PackRotamersMover will properly re-apply the TaskFactory.) We can also make the interface of PackRotamersMover a bit nicer


lyskov (approved):

PR #150, opened 2024-09-05
Fix issue with chain endings in cyclic peptide PDB reading.

roccomoretti:
@msklodow noticed that the InterfaceAnalyzer had an issue trying to score a cyclic peptide (chain B) bound to a protein (chain A). The reason for this is that the Rosetta PDB reader was basing chain ending annotation on terminal residue types. Since a cyclic peptide doesn't have a terminal residue, it was being "merged" with the prior chain (despite being a different chain letter).

This PR reworks how the chain ending annotation works. I think this generalizes things more properly. I'm not seeing issues with the expanded chain ending designations, but we can always tweak if there's issues.


lyskov (approved):
LGTM!

vmullig:
This could conceivably fix longstanding bugs with read-in of cyclic peptides from multi-chain PDB files.

PR #161, opened 2024-09-12
Remove is_surface() from Residue/ResidueType.

roccomoretti:
This function is confusing, as it's not talking about the surface of a protein, but mineral surfaces.

The ResidueType property SURFACE isn't used frequently -- for the rare case it's needed, we can use the more involved query protocol, rather than leaving the potentially confusing `Residue::is_surface()` lying around.


ajasja:
Looks good to me. (One option would be to rename to `is_type_surface()`)

PR #170, opened 2024-09-20
rename pytorch → torch extras and defines

lyskov:
looks like #43 missed a few things 🤦🏻‍♂️


MoritzErtelt (approved):
Thanks Sergey for fixing these!

PR #172, opened 2024-09-24
Building PHENIX interface - add missing imports 

folmilsch:
There are errors while building the PHENIX interface. I added the missing imports. These are not sorted according to "Project headers", "Utility headers" etc.


roccomoretti:
I'm guessing those headers disappeared because of the unused-header removal code. To avoid that happening in the future, I added a commit which wraps them in the same #ifdef used for the rest of the code. That should hopefully avoid them getting stripped out if we ever re-run the header code again.

roccomoretti (approved):
Should be good.

Note: CLA signature confirmed.

PR #181, opened 2024-09-30
Add debug printing for inline residue selector parsing logic.

roccomoretti:
There was some discussion (by e.g. @nannemdp) that the inline logic for ResidueSelectors doesn't always do the same thing as the separate AND/OR/NOT selectors do. This PR adds a rudimentary nested residue selector printing facility, and enables it with debug output on the core.select.residue_selector.util tracer when parsing inline logic.

It's primarily for debug purposes, and may or may not be useful in practice. (I didn't see any evidence of mis-parsing in the example under question.)

### DIFF
--- a/source/src/core/select/residue_selector/AndResidueSelector.cc
+++ b/source/src/core/select/residue_selector/AndResidueSelector.cc
@@ -171,6 +171,18 @@ std::string AndResidueSelector::class_name() {
 	return "And";
 }
 
+std::string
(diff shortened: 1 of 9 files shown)


nannemdp:
I don't know that they don't work in the same way, but that I can't get them to work at all. I know, it's unexplainable, and I don't have evidence in hand. But when I use explicit logic selectors I get the functionality I need...

PR #186, opened 2024-10-02
Fix typo

jvasile:
I fixed a quick typo in a comment, but... here's a question: has the decorator module been installed on the test server?  It might be time to remove the try/except training wheels (and the comment I fixed).


lyskov:
Thank you for spotting and fixing this @jvasile ! 

Re "decorator module installed...": - this comment is a bit misleading, reason for try/except block is not the absence of the decorator module on a testing server but rather our intention to make sure that PyRosetta can work without it. I will rephrase this comment shortly.

The Distributed framework (included in PyRosetta) require quite a number of additional packages along with special build setting for PyRosetta itself. So not all PyRosetta install will support that functionality hence the reason for try/except block there.

PR #197, opened 2024-10-10
Negative unit test for bad patches

JackMaguire:
I'm making a unit test out of [my recent post in the debugging channel](https://rosettacommons.slack.com/archives/C1YUVUD5E/p1728433183443519). It may not be the perfect error message but it is better than a seg fault and invites the user to expand upon the error message.

I wrote this code in the browser so I'm expecting a bit of beautification and typo correction to be necessary.

PR #204, opened 2024-10-14
Add saving of the beautification diff to the test server.

roccomoretti:
From a comment by @JackMaguire in #197, we should be able to save the full diff of the beautification test output, and present it on the benchmark server in the `Test files:  file-tree-view 」「 file-list-view 」` entries.


lyskov (approved):
I like this idea, - thank you @roccomoretti !!!

PR #221, opened 2024-10-23
Bugfix for RosettaCM and branched glycans

nikoniko90:
RosettaCM, with a branched glycan ligand, was segfaulting. In 2 locations, the "previous residue" of the given polymer entity was selected by choosing the position prior to the one in question, which is not appropriate for a branched glycan. The fix was to set a conditional for carbohydrates that uses conformation::carbohydrates::find_seqpos_of_saccharides_parent_residue to find the previous residue.

### DIFF
--- a/source/src/core/conformation/util.cc
+++ b/source/src/core/conformation/util.cc
@@ -71,2643 +71,2643 @@ static basic::Tracer TR( "core.conformation.util" );
 
 
 namespace core {
-	namespace conformation {
+namespace conformation {
 
-		void
-			orient_residue_for_ideal_bond(
-					Residue & moving_rsd,
-					chemical::ResidueConnection const & moving_connection,
-					Residue const & fixed_rsd,
-					chemical::ResidueConnection const & fixed_connection,
-					Conformation const & conformation,
-					bool lookup_bond_length
-					)
-			{
+void
+orient_residue_for_ideal_bond(
+	Residue & moving_rsd,
+	chemical::ResidueConnection const & moving_connection,
+	Residue const & fixed_rsd,
+	chemical::ResidueConnection const & fixed_connection,
+	Conformation const & conformation,
+	bool lookup_bond_length
+)
+{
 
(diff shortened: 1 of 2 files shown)


roccomoretti:
Please use the `../tools/python_cc_reader/beautify_changed_files_in_branch.py` script to adjust the formatting of the code -- right now it looks like your IDE reindented everything, which causes the PR diff to be rather extensive.

nikoniko90:
Thanks Rocco. Beautifier has been run.

source/src/core/conformation/util.cc
```diff
@@ -319,20 +320,36 @@ idealize_position(
...
+			if ( seqpos < conformation.size() && !rsd.is_upper_terminus() && !conformation.fold_tree().is_cutpoint( seqpos ) ) {
+				upper_connect = true;
+				core::Size child_res_seqpos( conformation::carbohydrates::find_seqpos_of_saccharides_mainchain_child( rsd ) );
+				ResidueOP next_rsd( ResidueFactory::create_residue( conformation.residue( child_res_seqpos ).type() ) );
+				idl.append_polymer_residue_after_seqpos( *next_rsd, idl_pos, true );
+			}
+		} else {
+			if ( seqpos > 1 && !rsd.is_lower_terminus() && !conformation.fold_tree().is_cutpoint( seqpos-1 ) ) {
+				lower_connect = true;
+				ResidueOP prev_rsd( ResidueFactory::create_residue( conformation.residue( seqpos-1 ).type() ) );
```
roccomoretti, line 341: The assumptions that a polymer is connected to `seqpos-1` & `seqpos+1` isn't necessarily valid, even in the case of proteins. -- But that's pre-existing code, so it shouldn't necessarily limit this PR. (Though if there was a generalization of UPPER/LOWER determination which works with carbohydrates, that might allow us to clean this up.)

roccomoretti (approved):
LGTM

PR #224, opened 2024-10-25
increase default unit-test-timeout to 8min

lyskov:
Recently we started to see frequent unit test timeouts, particularly when run on higher versions of GCC/Clang (probably more debug checks?). These create a false-negative reports which are destructive.

### DIFF
--- a/source/test/run.py
+++ b/source/test/run.py
@@ -722,7 +722,7 @@ def main(args):
     )
 
     parser.add_option("--timeout",
-      action="store",type="float", default=6,
+      action="store",type="float", default=8,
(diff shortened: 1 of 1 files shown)


roccomoretti (approved):
Would be better if we could figure out why the recent uptick, but this will also work, if we have the test server capacity.

lyskov:
> Would be better if we could figure out why the recent uptick, but this will also work, if we have the test server capacity.

@roccomoretti I think all failures are happening with GCC version 9 or later. So my theory is that these compilers produce debug builds that have some extra checks which slow things down. It would be great if we can get more definitive answers but i do not see any "actionable" items on this, - do you?

roccomoretti:
Sounds like a reasonable theory. If we're not crunched for test server time, we should be fine for bumping the limit.

PR #263, opened 2024-12-06
Lanthipeptide and recon multistate design update

tydingcw:
Modifications to lanthionine ring parameters to more accurately model the sidechain torsions and the methyl placement for methyllanthionine. Additionally, the recon multistate design code is updated to allow for design with NCAA by switching from a single letter code sequence stored in a string to a vector1 of strings that store the base name of the residue.


source/src/protocols/cyclic_peptide/crosslinker/lanthionine_util.cc
```diff
@@ -215,6 +215,10 @@ set_up_lanthionine_constraints(
 		pose.residue_type(dalares).atom_index("V1"),
 		dalares
 	);
+    AtomID const atom_bb_c(
```
roccomoretti, line 218: You need to run the beautifier prior to merging.

source/src/protocols/recon_design/FindConsensusSequence.cc
```diff
@@ -230,24 +230,41 @@ void FindConsensusSequence::apply_mpi( core::pose::Pose & pose ) {
 	utility::vector1< core::Size > my_designable_residues = get_designable_residues( pose, this_nodes_resfile );
 
 	/// Make a string out of the AAs at my designable positions in the current state
-	std::string my_sequence = get_designable_sequence ( pose, my_designable_residues );
+	utility::vector1< std::string > my_sequence = get_designable_sequence ( pose, my_designable_residues );
 
-	utility::vector1<std::string> all_pose_sequences (n_procs_);
+    std::string pass_seq;
+    for( const std::string& resi_base_name: my_sequence) {
```
roccomoretti, line 236: Rosetta convention is "east const": `std::string const &`

source/src/protocols/recon_design/MSDMover.cc
```diff
@@ -280,34 +281,50 @@ MSDMover::apply_mpi( core::pose::Pose & pose ) {
...
 		if ( rank == ii ) {
 			for ( core::Size jj = 1; jj <= n_procs; ++jj ) {
-				if ( rank!=jj ) utility::send_string_to_node( jj-1, my_sequence ); // node ranks are 0-indexed
+				if ( rank!=jj ) utility::send_string_to_node( jj-1, pass_seq ); // node ranks are 0-indexed
 				else other_pose_sequences[jj] = my_sequence;
 			}
 		} else {
-			other_pose_sequences[ii] = utility::receive_string_from_node( ii-1 ); // node ranks are 0-indexed
+            std::string passed_seq;
+			passed_seq = utility::receive_string_from_node( ii-1 ); // node ranks are 0-indexed
```
roccomoretti, line 302: Combine this on a single line. (In general, when possible, try to declare variables on the same line you first initialize them.)

source/src/protocols/recon_design/MSDMover.cc
```diff
@@ -280,34 +281,50 @@ MSDMover::apply_mpi( core::pose::Pose & pose ) {
...
+            //Need to split passed_seq by spaces
+            std::istringstream iss(passed_seq);
+            std::string resi_name;
+            while (iss >> resi_name){
+                other_pose_sequences[ii].push_back(resi_name);
+            }
```
roccomoretti, line 308: There's `split()`/`split_whitespace()` functions in `utility/string_util.hh` which does basically this.

source/src/protocols/recon_design/recon_util.cc
```diff
@@ -45,31 +45,40 @@ get_designable_residues( core::pose::Pose & pose, std::string resfile ) {
...
+        sequence.push_back(pose.residue_type( seqpos ).base_name());
 	}
 	return sequence;
 }
 
 /// Based on a list of sequences from poses, get all the AAs present at
 /// the position given by position_no
 utility::vector1< std::string >
-get_candidate_AAs( utility::vector1< std::string > other_pose_sequences,
+get_candidate_AAs( utility::vector1< utility::vector1< std::string > > other_pose_sequences,
```
roccomoretti, line 62: You probably want to pass unchanged vectors by `const &` rather than by value -- it saves data copies. (It's a pre-existing issue here, but is exasperated by adding the nesting and larger strings.)

source/src/protocols/recon_design/recon_util.cc
```diff
@@ -45,31 +45,40 @@ get_designable_residues( core::pose::Pose & pose, std::string resfile ) {
...
-std::string
+utility::vector1< std::string >
 get_designable_sequence ( core::pose::Pose & pose, utility::vector1< core::Size > designable_residues ) {
-	std::string sequence = "";
+	//std::string sequence = "";
+    utility::vector1< std::string > sequence;
 	for ( core::Size seqpos: designable_residues ) {
-		sequence += pose.residue( seqpos ).name1();
+		//sequence += pose.residue( seqpos ).name1();
+        sequence.push_back(pose.residue_type( seqpos ).base_name());
```
roccomoretti, line 54: One of the issues you'll run into here is that it looks like ResidueTypeConstraint takes a string which it compares with the `name3()` of the residue types. In general, the `base_name()` is not going to be equal to the `name3()`. (It is equal for standard residues, but isn't necessarily for noncanonicals and other residues which don't have a simple one letter code -> base_name() correspondence.)

(My recommendation is to either use `name3()` here, or add a mode to ResidueTypeConstraint which works on `base_name()`.)

source/src/protocols/recon_design/MSDMover.cc
tydingcw: If I use that fxn, is there any way to avoid this awkward passing the output to other_pose_sequences:

utility::vector1<std::string> split_seq = split_whitespace(passed_seq);
    for (std::string const &resi_name : split_seq) {
        other_pose_sequences[ii].push_back(resi_name);
    }

everyday847: You don't have to push it one string at a time; you should be able to set `other_pose_sequences[ ii ]` directly as the return value from `split_whitespace`.

roccomoretti (approved):

PR #277, opened 2024-12-16
Add cmath includes

roccomoretti:
`std::abs(Real)` needs the real-valued version of std::abs() function, which is found in the cmath header. (Versus the integer-valued versions, which can also be found in the cstdlib header.)

We were missing those imports in a few locations, which was causing my compiler to complain.


roccomoretti:
Can't say it's all the locations, but it's all the locations where the compiler was complaining.

PR #300, opened 2025-01-16
SID_ERMS_Rescore

robert-bolz:
This application builds on the existing SID_ERMS_prediction application (written by jtseffer). It simulates the ERMS data for an input structure, and rescores it based on the given experimental ERMS data. The rescored structures are given as the output (as well as the difference between the simulated and experimental ERMS data).


roccomoretti:
Looks like you've added an `inputs/` top level directory and left some convenience scripts in the branch. (It also looks like you have two `SID_ERMS_Rescore.cc` -- I don't know if you need both.) Could you clean those up?

robert-bolz:
I just cleared the top level directory and deleted the second ERMS_app.cc. Thanks!

source/src/apps/public/analysis/SID_ERMS_Rescore.cc
```diff
@@ -0,0 +1,673 @@
...
+		out << "Structure" << "\t" << "RMSE" << "\t" << "Ref15_Score" << "\t" << "ERMS Rescore" << std::endl;
+
+		utility::vector1<core::Real> rmse_list;
+		utility::vector1<core::Real> score_list;
+
+		//pose specific input checks
+
+		utility::vector1<std::string> files;
+		utility::vector1<file::FileName> list;
+		if ( option[in::file::l].user() ) {
```
roccomoretti, line 555: I'd probably recommend using `streams_from_cmd_line()` from core/import_pose/pose_stream/util.hh -- this parses the command line and loads the poses for you.  (And is flexible with various inputs: -s/-l/-in:file:silent/etc.)

```
MetaPoseInputStream input = streams_from_cmd_line();
core::pose::Pose current_pose;
while ( input.has_another_pose() ) {
    input.fill_pose( current_pose, *rsd_set );

    std::cout << "Read pose for file: " << core::pose::tag_from_pose(current_pose) << std::endl;
}
```

source/src/apps/public/analysis/SID_ERMS_Rescore.cc
```diff
@@ -0,0 +1,673 @@
...
+			}
+
+
+			//simulate ERMS
+			utility::vector1<utility::vector1<core::Real>> ERMS_prediction_(n_chains_);
+			simulate_ERMS(ERMS_prediction_, ACE_, B_, n_chains_, edges_, A_, breakage_cutoff_, nodes_);
+			TR << "ERMS simulation complete" << std::endl;
+
+			//calculte RMSE (if applicable)
+			if ( option[ RMSE ].user() && option[ ERMS ].user() ) {
```
roccomoretti, line 616: For boolean options, prefer `value()` to `user()`

(If someone sets `-RMSE false` on the command line, `user()` is true, but `value()` is false.)

source/src/apps/public/analysis/SID_ERMS_Rescore.cc
```diff
@@ -0,0 +1,673 @@
...
+					TR << "B values read in from file" << std::endl;
+					if ( option[ basic::options::OptionKeys::in::file::l ].user() ) {
+					TR.Warning << "PDB file input not used. B values read from file instead. To calculate B from structure, remove the flag (B_vals)." << std::endl;
+					}
+				} else {
+					if ( !option[ basic::options::OptionKeys::in::file::l ].user() ) {
+						std::stringstream err_msg;
+						err_msg << "Must input either file containing B values (using flag B_vals) or PDB file to calculate B values for simulation." << std::endl;
+					utility_exit_with_message(err_msg.str());
+				}
```
roccomoretti, line 587: You need to beautify -- that last curly brace isn't matched with the else clause, despite looking like it is from the indentation.

source/src/apps/public/analysis/SID_ERMS_Rescore.cc
```diff
@@ -0,0 +1,673 @@
...
+		for ( core::Size i = 1; i <= rmse_list.size(); i++ ) {
+        	if ( max_rmse < rmse_list[i] ) {
+            	max_rmse = rmse_list[i];
+        	}
+        }
+		for ( core::Size i = 1; i <= rmse_list.size(); i++ ) {
+        	if ( min_rmse > rmse_list[i] ) {
+            	min_rmse = rmse_list[i];
+        	}        
+        }
```
roccomoretti, line 646: If found it interesting you used std::max_element above, but you're not using it (or std::minmax_element) here. (That's fine, I just found it interesting.)

roccomoretti (changes requested):
Beautification before merging is a must.

Copy/paste code duplication is also something to be addressed - please move the common code into protocols to share between the applications. (Feel free to make a new protocols subdir if none of the current ones fit -- just be general in naming, such that future loosely related work can go into there as well.)

PR #303, opened 2025-01-20
Rosetta Evolutionary Ligand (REvoLd)

Hackefleisch:
This PR aims to include the newly developed application REvoLd into Rosetta.

REvoLd preprint: https://arxiv.org/abs/2404.17329

The algorithm optimizes small molecule ligands within combinatorial libraries like Enamine REAL. Most code changes are self enclosed and do not affect any functionality outside of the application. The only exception are changes to cmake compile directives for mpi which were broken.

A guide on how to run REvoLd is available here: https://docs.rosettacommons.org/docs/wiki/revold

Unit and integration test run locally and passed all.


source/src/apps/public/revold.cc
```diff
@@ -0,0 +1,69 @@
+// -*- mode:c++;tab-width:2;indent-tabs-mode:t;show-trailing-whitespace:t;rm-trailing-spaces:t -*-
```
roccomoretti, line 1: For organizational purposes, it might be worth putting the app in a subdirectory, rather than in the root applications dir. (Probably something suitably generic, like `ligand_design`.)

source/src/protocols/ligand_evolution/FragmentLibrary.cc
```diff
@@ -0,0 +1,645 @@
...
+	if ( reagents_.empty() ) {
+		TR.Error << "Unable to load reagents." << std::endl;
+		utility_exit_with_message( "Unable to load reagents." );
+	}
+}
+
+core::conformation::ResidueOP FragmentLibrary::create_ligand( std::string const& smiles, bool create_rotamers ) const {
+
+	// TODO Rosetta can't handle salts well, like an added Cl ion. Do I want to do something about that?
+	//      This is for now caught by the scorer and a score of 9999.9 is assigned to the molecule
```
roccomoretti, line 256: Given that physiologically most salts won't be associated with the ligand when bound, it's probably best to strip them out and use the version of the compound which would actually be pharmaceutically relevant. 

There's various "desalting" methods out there for RDKit (Though adding that is not necessary for this PR - I just mention it for completeness.)

source/src/protocols/ligand_evolution/FragmentLibrary.cc
```diff
@@ -0,0 +1,645 @@
...
+
+
+	RDKit::RWMol mutable_product( *RDKit::SmilesToMol( smiles ) );
```
roccomoretti, line 259: This is making a copy of the returned object. Additionally, as SmilesToMol() returns a raw pointer, you're leaking memory -- I'd recommend sticking the return value in a RDKit::RWMOL_SPTR (or the convenience typedef RDKit::RWMolOP ) that should fix memory issues.

source/src/protocols/ligand_evolution/FragmentLibrary.cc
```diff
@@ -0,0 +1,645 @@
...
+
+	RDKit::MolOps::sanitizeMol( mutable_product );
+	RDKit::MolOps::addHs( mutable_product );
+
+	RDKit::DGeomHelpers::EmbedMolecule( mutable_product );
```
roccomoretti, line 264: Just a quick note that the version of RDKit we currently have in main has somewhat sub-optimal molecule embedding by default. (Notably, there's some basic knowledge items which greatly improve the geometry of generated conformers.)

I'd recommend passing the `RDKit::DGeomHelpers::srETKDGv3` parameter set as the second parameter to get more sensible ligand structures.

source/src/protocols/ligand_evolution/FragmentLibrary.cc
```diff
@@ -0,0 +1,645 @@
...
+	utility::vector1< std::map< std::string, core::Vector > > return_value;
+
+	core::chemical::rdkit::RestypeToRDMol to_rdmol( new_ligand, /*neutralize = */ false, /*keep_hydro=*/ true );
+	::RDKit::RWMOL_SPTR rdmol( to_rdmol.Mol() );
+
+	core::chemical::VDIndexMapping const & rosetta_to_rdkit( to_rdmol.vd_to_index() );
+
+	// Generate the rotamers
+	::RDKit::INT_VECT conf_ids; // List of generated conformations
+	::RDKit::DGeomHelpers::EmbedMultipleConfs(*rdmol, conf_ids, nconf,
```
roccomoretti, line 309: The funky signature here was to set the basic knowledge parameters with an older version of RDKit -- Again, srETKDGv3 is probably the easier way to go.

source/src/protocols/ligand_evolution/FragmentLibrary.cc
```diff
@@ -0,0 +1,645 @@
...
+
+	core::conformation::ResidueOP tmp_residue( new core::conformation::Residue( restype, true ) );
+
+	TR.Debug << "Done creating." << std::endl;
+
+	return tmp_residue;
+}
+
+core::Size FragmentLibrary::generate_rotamers( core::chemical::MutableResidueType& new_ligand ) const {
+	// This whole function is written by and copied from Rocco Moretti !
```
roccomoretti, line 284: If you're only using it with the function above, it might be more efficient to embed all conformers from the original RWMol, then convert, then pull out the conformers into the StoredRotamerLibrarySpecification, rather than bouncing back-and-forth between the representations.

source/src/protocols/ligand_evolution/FragmentLibrary.cc
```diff
@@ -0,0 +1,645 @@
...
+Reagent::Reagent( std::string const& name, std::string const& reagent_smiles )
+:
+	name_( name ),
+	mol_( RDKit::SmilesToMol( reagent_smiles ) )
+{
+	fingerprint_ = RDKit::MorganFingerprints::getFingerprint( *mol_, 2 );
+}
+
+Reagent::~Reagent() {
+	delete fingerprint_;
```
roccomoretti, line 641: Probably better to use a `std::shared_ptr` (or `std::unique_ptr`) versus using manual memory management.

source/src/protocols/ligand_evolution/FragmentLibrary.cc
```diff
@@ -0,0 +1,645 @@
...
+	TR.Debug << "Running reaction: " << reactions_[ reaction_index ]->name_ << " with";
+	for ( core::Size ii( 1 ); ii <= n_reaction_positions; ++ii ) {
+		TR.Debug << " reagent" << ii << " " << reagents_[ reagent_indices[ ii ] ]->name_;
+	}
+	TR.Debug << std::endl;
+
+	// rdkit requires some weird vector and vector of vector shenanigans
+	RDKit::MOL_SPTR_VECT react;
+	for ( core::Size ii( 1 ); ii <= n_reaction_positions; ++ii ) {
+		react.push_back( RDKit::RWMOL_SPTR( reagents_[ reagent_indices[ ii ] ]->mol_ ) );
```
roccomoretti, line 508: `reagents_[ reagent_indices[ ii ] ]->mol_` should already be a RDKit::RWMOL_SPTR, right?

source/src/protocols/ligand_evolution/FragmentLibrary.cc
```diff
@@ -0,0 +1,645 @@
...
+	}
+
+	auto products = reactions_[ reaction_index ]->reac_->runReactants( react );
+	TR.Debug << "Len of outer vector: " << products.size() << std::endl;
+	TR.Debug << "Len of inner vector: " << products[ 0 ].size() << std::endl;
+
+	return RDKit::MolToSmiles( RDKit::RWMol( *products[ 0 ][ 0 ] ) );
```
roccomoretti, line 515: Is there a need to make a copy of the RWMol here? MolToSmiles should be able to work on the dereferenced ROMol directly.

source/src/protocols/ligand_evolution/FragmentLibrary.hh
```diff
@@ -0,0 +1,231 @@
...
+	core::Size n_positions() const;
+
+private:
+
+	std::string name_;
+	utility::vector1< utility::vector1< core::Size > > reagents_;
+	RDKit::ChemicalReaction* reac_;
+	core::Size possible_molecules_ = 1;
+
+	friend FragmentLibrary;
```
roccomoretti, line 225: Try to avoid `friend` if at all possible. It took me way to long to understand where/how `reagents_` was getting set -- simply having an "add_reagent()" function might better help encapsulate things and make the flow clearer.

source/src/protocols/ligand_evolution/offspring_factory/OffspringFactory.cc
```diff
@@ -0,0 +1,22 @@
...
+
+/// @file   protocols/ligand_evolution/offspring_factory/OffspringFactory.cc
+/// @brief  The (mostly empty) definition of the %OffspringFactory class
+/// @author Paul Eisenhuth (eisenhuth451@gmail.com)
+
+// unit headers
+#include <protocols/ligand_evolution/offspring_factory/OffspringFactory.hh>
+
+
+namespace protocols {
```
roccomoretti, line 18: I'll note that you don't need to have a cc file if there's nothing to put in it. (Especially pure virtual classes can be header-only.)

roccomoretti (commented):
Looks rather good -- Some minor comments, but should be more-or-less good to go once testing fully passes.

source/src/protocols/ligand_evolution/EvolutionOptions.hh
```diff
@@ -0,0 +1,294 @@
...
+	std::string path_to_reactions_;
+
+	/// @brief Path where to find reagents definitions. The standard settings assume you use a input and run directory structure.
+	std::string path_to_reagents_;
+
+	bool selectors_defined_ = false;
+
+	/// @brief Maps selector names to types and parameter list
+	std::map< std::string, std::pair< std::string, std::map< std::string, core::Real > > > selector_options_ {
+{ "remove_elitist", { "elitist", {
```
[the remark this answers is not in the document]
Hackefleisch: This was indented, but the beautifier squashed everything left. Is there a way to prevent it from doing so?

source/src/apps/public/revold.cc
Hackefleisch: Done

source/src/apps/public/revold.cc
```diff
@@ -0,0 +1,69 @@
...
+		int size = 1;
+#ifdef USEMPI
+        MPI_Init (&argc, &argv);
+        MPI_Comm_rank (MPI_COMM_WORLD, &rank);
+        MPI_Comm_size( MPI_COMM_WORLD, &size );
+
+        if ( size < 10 ) {
+            TR.Warning << "Running REvoLd less than 10 CPUs causes very large runtimes. Consider using more CPUs." << std::endl;
+        }
+#else
```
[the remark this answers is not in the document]
Hackefleisch: Done

source/src/protocols/ligand_evolution/EvolutionManager.cc
```diff
@@ -0,0 +1,404 @@
...
+
+void EvolutionManager::score() {
+#ifdef USEMPI
+        work_manager_->score( population_ );
+#else
+	scorer_->score_population( population_ );
+#endif
+}
+
+std::string EvolutionManager::print_scores() const {
```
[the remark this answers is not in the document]
Hackefleisch: Changed it to scores_to_str()

source/src/protocols/ligand_evolution/EvolutionManager.cc
```diff
@@ -0,0 +1,404 @@
...
+	core::Size smiles_to_score = library_.n_unscored_smiles() / mpi_size;
+	for ( core::Size ii( 1 + smiles_to_score * rank_ ); ii <= smiles_to_score * ( rank_ + 1 ); ++ii ) {
+		try {
+			scorer_->score_ligand( { ii, 0 }, "", external_scoring_ );
+			if ( ii % 10 == 0 ) {
+				scorer_->save_external_scoring_results( core::Size( rank_ + 1 ));
+			}
+		} catch( ... ) {
+			// print message and continue scoring
+			TR.Error << "Unable to score " << library_.identifier_to_smiles( { ii, 0 } ) << std::endl;
```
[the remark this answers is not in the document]
Hackefleisch: Good catch! I am assuming that this try catch block was a hacky workaround before I caught exceptions during scoring specifically which handles setting the score. I remove this try catch block here.

source/src/protocols/ligand_evolution/FragmentLibrary.cc
Hackefleisch: I gave this another look and found only 16 fragments containing a salt ion in the combinatorial library. The "failed" scores was only assigned five times during the final benchmark docking a total of ~360k molecules, so I think it is really save to deem this not important. And lastly, the desalting methods for RDKit are only available through python: https://github.com/rdkit/rdkit/issues/700

source/src/protocols/ligand_evolution/FragmentLibrary.cc
Hackefleisch: Done

source/src/protocols/ligand_evolution/FragmentLibrary.cc
Hackefleisch: Done

source/src/protocols/ligand_evolution/FragmentLibrary.cc
Hackefleisch: Done

source/src/protocols/ligand_evolution/FragmentLibrary.cc
Hackefleisch: Added this as a todo comment to the code

source/src/protocols/ligand_evolution/FragmentLibrary.cc
Hackefleisch: Done

source/src/protocols/ligand_evolution/FragmentLibrary.cc
Hackefleisch: You are right and I removed that unneccessary copy ctor

source/src/protocols/ligand_evolution/FragmentLibrary.cc
Hackefleisch: Again correct, copy ctor is removed now

source/src/protocols/ligand_evolution/FragmentLibrary.hh
Hackefleisch: Done

source/src/protocols/ligand_evolution/offspring_factory/OffspringFactory.cc
Hackefleisch: Done

roccomoretti:
While one of the scripts just checks the files which changed, this is just for user convenience (shorter runtime). The script on the test server runs through all of the files (including the ones which haven't changed in that commit.)

PR #306, opened 2025-01-22
fix Conda release procedure

lyskov:
Recently Conda changed extension suffix for its packages which broke our release procedure. This PR is intended to fix this.


roccomoretti (approved):
Is this something where we need to potentially support previous versions of Conda? Does the new way work for all relevant Conda versions, or should we have some sort of if/then/else difference in behavior based on the Conda version being used?

lyskov:
> Is this something where we need to potentially support previous versions of Conda? Does the new way work for all relevant Conda versions, or should we have some sort of if/then/else difference in behavior based on the Conda version being used?

-- Good point Rocco! Since we are re-deploying Conda environment each time then we automatically will get the latests version of Conda build tools installed.

Technically it was possible to to make our script work with any version by simply removing suffix from regex and instead calculate it each time based on what Conda build script prints 🤔. It kind of neat but at the same time i do like extra "sanity check" so to speak so our scripts "fails loudly" if output became unexpected. Let me think about this 🧐, maybe after all this is a way to go…

PR #349, opened 2025-03-05
FakeRotLib and minor tweaks to m2pp.py

ewbell94:
This pull request is to introduce the FakeRotLib script formally into the rosetta codebase.  The script takes in a NCAA sdf, generates conformers using RDKit, parameterizes it using molfile_to_params_polymer.py, and optionally creates a rotlib file describing the rotamer distributions of the chi torsions.

A few small tweaks were also made to molfile_to_params_polymer.py to ensure that the backbone hydrogen is always described using a proper torsion and not an improper torsion and to ensure guanadinium nitrogens are properly typed.


source/scripts/python/public/fake_rotlib.py
```diff
@@ -0,0 +1,524 @@
...
+    #Generate dipeptide form and RosettaParams instructions
+    if args.dip:
+        try:
+            ncaamol = Chem.MolFromMolFile(args.input, removeHs=False)
+            if ncaamol == None:
+                raise Exception("Molecule read-in returned None.")
+        except Exception as e:
+            print("ERROR reading in SDF file:",e)
+            exit(1)
+        instructions = generateInstructions(ncaamol, args.n_cbb)
```
tydingcw, line 474: Note that Chem.AssignStereochemistryFrom3D(mol) is called in the geomOptimize function. This function is not called if you use the --dip option. So steriochemistry is not assigned with --dip and this can lead to issues where rdkit forgets what the sterichemistry should be.

roccomoretti (approved):
I was going to recommend adding an integration test for standard usage, such that we test that things run, but that would only work if rdkit is installed on the test system's python, which we can't guarantee.

PR #356, opened 2025-03-12
Better debugging on CIF file read in.

roccomoretti:
Issue #355 indicates that errors in mmCIF parsing are not well handled. This commit attempts to fix that in two ways.

The first is to add a 'from extension' fallback if the from-contents parser doesn't like it. We assume that if you're naming it xyz.pdb, you want it parsed as a PDB.

The second is to print the diagnostics report for mmCIF files to the tracer -- this is what's being used to autodetect mmCIF format, so better transparency of what the failure issue is would be warranted.


lyskov (approved):
LGTM!

PR #372, opened 2025-03-24
Add option to control intermediate dumping in BackrubProtocol

roccomoretti:
Especially with XML (where the Backrub protocol is part of a more extensive protocol),  you don't necessarily want an unconditional dumping of the last & low results from the Backrub stage.  This PR adds an XML option which will allow the dumping of the poses (old behavior), but turns it off by default.  This only changes XML usage -- I didn't change how the backrub application behaves.

 Changes to the backrub_interface_ddG integration test is expected.

### DIFF
--- a/source/src/protocols/backrub/BackrubProtocol.cc
+++ b/source/src/protocols/backrub/BackrubProtocol.cc
@@ -101,39 +101,7 @@ BackrubProtocol::BackrubProtocol():
 	read_cmd_line_options();
 }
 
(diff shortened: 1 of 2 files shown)


woodsh17 (approved):
@roccomoretti Do you know if there are any tests that use BackrubProtocol outside of RosettaScripts to make sure that behavior stays the same? Also, do we need to test that if dump_poses_ is set to true that the PDBs are in fact dumped?

PR #374, opened 2025-03-26
Fix for the maintenance.documentation test

roccomoretti:
With the new open repository, direct committing to the repos don't really work anymore.

Instead, we can have the test simply test for changes, and provide a patch file that can be applied to the documentation repository.

This PR also updates the documentation repo to pull in all the changes made through the wiki interface since the last update.


lyskov (approved):
LGTM!

PR #375, opened 2025-03-27
Fix import of load_ligand.py

roccomoretti:
Fix issues with `import pyrosetta.toolbox.load_ligand` -- This won't fix things for people who use the functions in question, but it will allow for use of the other functions in this module.

Thanks to @decortja for bringing this to my attention.


lyskov (approved):
LGTM, - thank you Rocco!

PR #388, opened 2025-04-08
fix Mac RC release package

lyskov:
Current Issue: Our Mac binary releases are currently built using the instruction set of the local Intel Xeon CPU (via -march=native), which can result in binaries that fail to run on other x86 CPUs due to unsupported instructions.

Fix: This PR addresses the issue by explicitly setting the instruction set to core2 when compiling on macOS with an x86 CPU, ensuring the resulting binaries are compatible with a wider range of Intel-based Macs.


roccomoretti (approved):

lyskov:
Thank you @roccomoretti !

PR #392, opened 2025-04-11
Revold

Hackefleisch:
This is a small change to address the failed clang-10.clang.cxx11thread.mpi.serialization.tensorflow.build.debug test.

The one line change removes a faulty precedence order of logical operators and simplifies the query in question. 

The same result could be achieved by assigning the int return value directly to the boolean, but I decided in favor of the comparison to increase readability.


lyskov (approved):
LGTM!

PR #406, opened 2025-05-07
fix Mac build on Clang-17

lyskov:
Thanks to @rclune for reporting this issue!

### DIFF
--- a/source/src/core/grid/CartGrid.hh
+++ b/source/src/core/grid/CartGrid.hh
@@ -459,27 +459,27 @@ public:
 
 	}
 
-	void sum(utility::vector0<utility::pointer::shared_ptr<CartGrid<T> > > const & list_grids) {
-		core::Size ngrids = list_grids.size();
-
-		this->zero();
-		T curr_value, new_value;
-		for ( core::Size i=0; i < ngrids; i++ ) {
-			if ( !this->equalDimensions(*(list_grids[i])) ) {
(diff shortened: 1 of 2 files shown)


source/src/core/grid/CartGrid.hh
```diff
@@ -459,27 +459,27 @@ public:
...
+	// 		}
+	//
+	// 		for ( core::Size j=0; j < this->its_npoints; j++ ) {
+	// 			curr_value = this->getValue(j);
+	// 			new_value = list_grids[i]->getValue(j);
+	// 			curr_value += new_value;
+	// 			this->setValue(j,curr_value);
+	// 		}
+	// 	}
+	// }
```
lyskov, line 483: @decarboxy tagging you here since you was listed as original author of this code.

-- @roccomoretti @decarboxy this is and interesting case of "dead" code. I was trying to figure out why this suddenly became broken (`its_npoints` member missing) when I realized that this was never used before. For some reason none of the previous compilers (and not even Clang-17 under Linux) tried to instantiate this function but new Apple Clang-17 did 🤔

roccomoretti (approved):
With the caveat that the current code breaks the beautifier.

Seems like the code is indeed unneeded. As such, deleting it should be fine.

lyskov:
@Thank you for review @roccomoretti ! I agree, lets delete it instead.

PR #419, opened 2025-05-19
Updates to molfile_to_params_polymer.py to better accomodate non-alpha residues.

roccomoretti:
This is PR RosettaCommons/main#6462 which was already approved by @nannemdp 

The current molfile_to_params_polymer.py is not suited to creating params file for non-alpha amino acids, but I was able to make some tweaks to generalize things.

Critically, there's now a M POLY_BACKBONE setting which can take the place of the alpha-amino-acid specific designations. You just need the ROOT, POLY_UPPER, POLY_LOWER and POLY_BACKBONE for things to work. -- Well, assuming you now use mol2 input and the new --keep_names flags, to avoid the alpha-amino-acid specific renaming code.


woodsh17 (approved):
My understanding is if you have a non-alpha amino acid, then you need to use mol2 input and set the --keep_names to true. Where should we document this? I saw David linked this in the previous PR: https://docs.rosettacommons.org/docs/latest/rosetta_basics/non_protein_residues/Noncanonical-Amino-Acids. Is there anywhere else?

PR #425, opened 2025-05-29
Updates to cryoEM ligand identification in GALigandDock

agmuenks:
This PR updates values in a model used to identify ligands in cryoEM data. The only changes are hard coded values in GALigandDock and the calculation method for a couple of variables.


source/src/protocols/ligand_docking/GALigandDock/GALigandDock.cc
```diff
@@ -602,12 +603,12 @@ GALigandDock::apply( pose::Pose & pose )
 					TR << "No local resolution provided. Using the global resolution" << std::endl;
 				}
 
-				core::Real expected_dens = 0.4535172 - 0.0190373 * local_res_ + 0.5542869 * pose_cc  -0.0006722 * nheavyatoms;
+				core::Real expected_dens = 0.3428443 - 0.0372441 * local_res_ + 1.1934549 * pose_cc  -0.0014546 * nheavyatoms;
 
-				core::Real dG_z = ((dG - expected_dG)/15.8 * -1 * 1.5);
-				core::Real dens_z = ((penalized_density - expected_dens)/(0.07192 * 0.6));
+				core::Real dG_z = ((dG - expected_dG)/15.8 * -1 * 0.8);
+				core::Real dens_z = ((atanh(penalized_density) - expected_dens)/(0.1531 * 1.1));
```
roccomoretti, line 609: I'd recommend importing cmath (versus math.h) and then explicitly using `std::atanh` -- While it probably doesn't matter for intrinsically floating point functions like `atanh()`, mixing plain C and C++ functions can be fraught for other functions (we've had issues with `abs()`). Using the C++ versions is more robust to type overloading.

roccomoretti (approved):
Looks fine to me.

I'd probably recommend leaving comments about where the magic numbers come from, such that people reading it in the future know why you went with 0.8 instead of 1.5 (for example)

(Extracting things off into configurable parameters is probably an even better idea, but not something that's strictly necessary before merging.)

source/src/protocols/ligand_docking/GALigandDock/GALigandDock.cc
agmuenks: Thanks for the suggestions Rocco! I added comments and changed to cmath in the most recent commit.

PR #456, opened 2025-07-02
Add RosettaLigand docking scores to the Pose when JD2 is not active.

roccomoretti:
People running RosettaLigand jobs under PyRosetta keep having problems that the computed docking scores don't show up. This is because the current RosettaLigand code was written for RosettaScripts and assumes the use of JD2. The scores get stuck in a JD2 job, which is hard to access when you're not running a command line JD2 job.

We can change the code to stick the scores into the pose instead of (in addition to) the job when not running under JD2 (e.g. PyRosetta) Hopefully this should make the access easier.


source/src/protocols/ligand_docking/util.hh
```diff
@@ -60,6 +60,11 @@ void move_ligand_neighbor_to_desired_position(
 	core::pose::Pose & pose
 );
 
+/// @brief Annotate a string/real pair.
+/// Will place it either in the JD2 Job (if JD2 active -- this is the historical behavior) or into the Pose
```
Hackefleisch, line 64: This comment is currently misleading as it will always place the score in the job and only in the pose if jd2 is not active.

Hackefleisch (commented):
Looks good to me. I would only prefer a general behavior of always adding the score to both job and pose if there are no strong reasons to no do so.

Hackefleisch (approved):
Looks good!

PR #458, opened 2025-07-03
Revold Integration Test Fix

Hackefleisch:
The REvoLd integration tests frequently broke due to time limitations. To fix this I did two things:

1. The integration tests are no longer one monolithic command file but is split into three independent tests.
2. All options and parameters were tweaked to allow short runtimes whilst accepting less accurate results (as accuracy doesn't matter for integration tests)

This shortened the runtime of the longest command to a third of the previous combined command.

Additionally, I gzipped the input pdb to reduce memory requirements.


nannemdp:
So glad to see that the integration tests will run reliably.

Can you make sure to comment what the ideal options are for a production run? Either in an options file as part of the test or in some application documentation. I often use the integration tests as an example of how to run applications or code and it can be a challenge to re-optimize the parameters.

I guess documenting the parameters in the manuscript (SI probably) would be okay, especially if there is a proper citation printed at the end of the run.

roccomoretti (approved):

PR #459, opened 2025-07-04
Add support for filtering results in PyRosettaCluster

klimaj:
Similar to the idea of filters in RosettaScripts, it can be useful to filter a result mid-trajectory in `PyRosettaCluster` (rather than solely post hoc filtering data). This PR supports optionally filtering results between user-defined PyRosetta protocols in `PyRosettaCluster`. 

Currently, when a user-defined PyRosetta protocol returns or yields `None` (e.g., a `PackedPose` was discarded in a protocol), `PyRosettaCluster` converts it to an empty `PackedPose` object that gets passed to the next protocol. This can result in needless additional overhead in `dask` because the empty `PackedPose`s still get distributed to all downstream PyRosetta protocols -- even if the protocols don't operate on empty input `PackedPose`s. The user usually has to catch these at the top of their protocols (e.g., with `if packed_pose.pose.empty(): return None`). This PR supports a new `PyRosettaCluster` instance attribute, `filter_results`, that dynamically filters out any empty `PackedPose`s that arrive back to the head node. If enabled, this is used for filtering out decoys mid-trajectory between user-provided PyRosetta protocols. 

Because this changes `PyRosettaCluster` behavior, if there are any users who happen to rely on these empty `PackedPose`s to generate results in downstream protocols (which are user-defined, so there may be scripts in the wild that depend on it not filtering), then decoy IDs in downstream protocols would be changed if enabled. Hence, for reproducibility of existing data with backward compatibility, this PR sets the default to `False` (although users ought to be implementing an identical PyRosetta version when reproducing results). Additionally, this PR ensures that the `filter_results` instance attribute cannot be overridden during reproduction. Hence, this is a new feature that can optionally be enabled for reduced overhead.

### DIFF
--- a/source/src/python/PyRosetta/src/pyrosetta/distributed/cluster/converter_tasks.py
+++ b/source/src/python/PyRosetta/src/pyrosetta/distributed/cluster/converter_tasks.py
@@ -422,8 +422,16 @@ def _parse_dict(obj: Dict[Any, Any]) -> Dict[Any, Any]:
     for k in obj.keys():
         if k in ["client", "clients", "input_packed_pose"]:
             raise NotImplementedError(
-                f"The parameter '{k}' must be passed directly to reproduce(), "
+                f"The parameter '{k}' must be passed directly to `reproduce()`, "
                 + "not as a member of the 'instance_kwargs' dictionary."
+            )
+        elif k == "filter_results":
+            raise ValueError(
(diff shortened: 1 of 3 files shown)


lyskov (approved):
LGTM!

lyskov (changes requested):
👍

lyskov (approved):

PR #431, opened 2025-06-02
Enhance pyrosetta and pyrosetta.distributed I/O interfaces

klimaj:
This PR adds convenience functions (with unit tests) to the `pyrosetta.io` and `pyrosetta.distributed.io` namespaces, in particular promoting buried functions like `pyrosetta.rosetta.core.io.pdb.dump_multimodel_pdb` to the top-level namespace like `pyrosetta.dump_multimodel_pdb`:

Upgrades to the `pyrosetta` and `pyrosetta.io` namespaces:
- Input functions (i.e., files to `Pose` objects):
  - Defines `pose_from_pdbstring`
  - Defines `poses_from_multimodel_pdb`
- Output functions (i.e., `Pose` objects to files):
  - Imports `dump_pdb`
  - Defines `dump_scored_pdb`
  - Imports `dump_cif`
  - Imports `dump_mmtf`
  - Defines `dump_multimodel_pdb`
  - Defines `dump_file`
- Score functions:
  - Imports `create_score_function` in `pyrosetta.io` namespace
  - Imports `get_fa_scorefxn` in `pyrosetta.io` namespace
  - Imports `get_score_function` in `pyrosetta.io` namespace

Upgrades to the `pyrosetta.distribued.io` namespace:
- Input functions (i.e., files to `PackedPose` objects):
  - Refactors `poses_from_files`
  - Defines `pose_from_pdb` (thanks to @ajasja for the suggestion to combine `pose_from_pdb_bz2`/`pose_from_pdb_gz` functions together, and @lyskov for suggesting support for `xz` compression -- this one function now handles all 3 compression formats)
  - Defines `pose_from_base64`
  - Defines `pose_from_pickle`
  - Defines `poses_from_sequences`
  - Defines `poses_from_multimodel_pdb`
- Output functions (i.e., `Pose` or `PackedPose` objects to files):
  - Defines `dump_pdb`
  - Defines `dump_scored_pdb`
  - Defines `dump_cif`
  - Defines `dump_mmtf`
  - Defines `dump_multimodel_pdb`
  - Defines `dump_pdb` (@ajasja @lyskov this one function supports `bz2`, `gzip` and `xz` compression)
  - Defines `dump_base64`
  - Defines `dump_pickle`
  - Defines `dump_file`

- Conversion functions:
  - Defines `pose_result`
  - Defines `PackedPose.clone()` method
  - Defines `to_base64`
  - Defines `to_pickle`

[...]

PR #462, opened 2025-07-05
Support saving PackedPose objects and arbitrary score data in PyRosettaCluster

klimaj:
This PR adds support for `PyRosettaCluster` to output decoys in the `.pose` file format (introduced in #431), as well as to output scorefiles in pickled `pandas.DataFrame` format (which supports arbitrary datatypes that have been saved in the `PackedPose.scores` dictionary as introduced in #430). 

Currently, `PyRosettaCluster` instance keyword arguments are saved on `REMARK` lines in output PDB files, and in this PR `PyRosettaCluster` instance keyword arguments are technically cached in the `pyrosetta.rosetta.core.pose.datacache.CacheableDataType.STRING_MAP` datacache, which results in a `REMARK` line if output as a PDB file, yet we save it as a `.pose` file. Therefore, this PR supports reproducing a decoy from a `.pose` file that has been output by `PyRosettaCluster`. The benefit of saving decoys in the `.pose` format over PDB format is that arbitrary python types are still cached in the `.pose` file and do not require JSON-encoding (since JSON-encoding can't serialize arbitrary data types). Additionally, `.pose` files save the exact atomic coordinates of the pose, and therefore it can be demonstrated that a decoy can be recursively reproduced _ad infinitum_ through a set of custom PyRosetta protocols with exact atomic coordinates (note that `PyRosettaCluster` has already been doing this, but because the PDB file format rounds the atomic coordinates it was not possible to explicitly show this).

Furthermore, this PR supports saving `PyRosettaCluster` scorefiles as pickled `pandas.DataFrame` files, which also can store arbitrary python types from the `PackedPose.scores` dictionary. Decoys can also be reproduced from a scorefile in pickled `pandas.DataFrame` format (which supports many additional compression types, including `.gz`, `.bz2`, `.xz`, `.tar`, `.tar.gz`, `.tar.xz`, and more). Please note that `pandas` is already a dependency of `dask`, and is therefore already a dependency of `pyrosetta.distributed`, so new third-party dependencies are not being added with this PR.  

This PR also adds a slight guardrail when attempting to reproduce decoys with a PyRosetta version that is not exactly identical to the version used to produce the original decoy. Currently, `PyRosettaCluster` just gives a warning if the PyRosetta versions are not identical between the build that created the original decoy and the current build being implemented for reproduction. In this PR, if there's a mismatch we raise an error instead, and the user must manually input an empty string to the `pyrosetta_build` keyword argument to bypass PyRosetta build validation. This is similar logic as the conda environment validation that is already in place.

Note: this PR can be merged after merging #430, since there is some overlapping code.


lyskov (approved):
LGTM!

I am a bit hesitant to "claim" `.pose` for extension though (it kind of imply that it is a Rosetta C++ format), - Jason how do you feel about maybe using `.pypose` or something similar? Ie to imply that it is a PyRosetta format after all.

@roccomoretti tagging you in case you have an opinion on this. Thanks,

source/src/python/PyRosetta/src/pyrosetta/distributed/cluster/core.py
```diff
@@ -118,8 +119,11 @@
     pyrosetta_build: A `str` or `NoneType` object specifying the PyRosetta build as
         output by `pyrosetta._version_string()`. If `None` is provided, then PyRosettaCluster
         automatically detects the PyRosetta build and sets this attribute as the `str`.
-        If a `str` is provided, then validate that the input PyRosetta build is equal
-        to the active PyRosetta build, and log a warning message if not.
+        If a non-empty `str` is provided, then validate that the input PyRosetta build is
+        equal to the active PyRosetta build, and raise an error if not. This ensures that
+        reproduction simulations use an identical PyRosetta build from the original
+        simulation. To bypass PyRosetta build validation with a warning message, an
+        empty string ('') may be provided (but does not ensure reproducibility).
```
lyskov, line 126: @klimaj what value exactly is used when checking if PyRosetta version is equal? Is that output of `pyrosetta._version_string()`? I am asking because technically that not enough to guarantee that versions are equal. More precise would be to use both version string and output of `pyrosetta.rosetta.utility.Version.extras()` (probably sorted and concatenated into a string). That way we can distinguish slightly different build flavors that could yield difference down the line, like builds with different extras.

lyskov (approved):
LGTM!

source/src/python/PyRosetta/src/pyrosetta/distributed/cluster/core.py
klimaj: @lyskov Oh good catch, thanks! Indeed we are just comparing `pyrosetta._version_string()` here (which is probably enough to distinguish the `pyrosetta_installer` PyPI package options `serialization=True/False` (which puts/omits `cxx11thread.serialization` to/from the `pyrosetta._version_string()`) and the `type='Release'/'MinSizeRel'` (also in the `pyrosetta._version_string()`). Still that's correct- I suppose a user could also do a custom PyRosetta install into a conda environment with some extras that would not be caught. So yes, I agree, perhaps we can add a `pyrosetta._complete_version_string()` or similar in PyRosetta (in a separate PR, since it's a bit out-of-scope of this PR)

PR #480, opened 2025-07-25
Fix loading of ligands when three letter code matches NCAA

roccomoretti:
@LAnAlchemist noticed that when a ligand params file provided with `-extra_res_fa` has a three letter code which matches an NCAA three letter code from the database, that ligand ResidueType is never selected on PDB read-in, even if it's a much better match for the names in the PDB.

The reason for this is that the PDB reader residue typer prioritizes patched polymeric terminus types (those with TERMINUS properties) for residues at the ends of chains, discarding the ligand types as a possibility before even encountering the name-based selection.

This PR adjusts how the typer selects residues. Instead of having chain-terminal residues preferring terminus properties, actually look at the connection points, and look for residues which have/don't have the UPPER & LOWER connection points. (This is really what "is_lower_terminus" and "is_upper_terminus" in PoseFromSFRBuilder signifies: is this residue polymerically connected to the adjacent residue.)


woodsh17 (commented):
Are the error messages popping up in integration tests rna_denovo_dna_bridge, swm_dna_bridge, posttranslationalmod_io expected from this change? The tests look like they still end the same though.

lyskov (approved):
Code LGTM!

roccomoretti:
Anticipated (debugging-level) tracer changes in integration tests (but no spurious error messages anymore).  Functionality changes in simple_metrics_b_factor integration test due to different flags,

The phosphonate integration test changes, but that's because an N-acetylation present in the input is now being correctly recognized.

PR #484, opened 2025-07-26
Set correct scope for automatic protocol detection in PyRosettaCluster reproduction

klimaj:
If using the `pyrosetta.distributed.cluster.reproduce` function and a list of user-defined PyRosetta protocols is not provided, then it attempts to automatically detect the protocols from the current stack frame. However, the function is naively searching in `globals()` from `tools.py`, and thus only searches for the protocols in the global variables of `tools.py` and not from the user's script. This PR fixes the automatic detection by instead searching up two frames to the caller of the `pyrosetta.distributed.cluster.reproduce` function. 

Furthermore, this PR extends functionality to search for protocols in both `globals()` and `locals()` of the user's script. In this way, whether the user imports/defines their protocols in the global or local variables, the protocols are detected.

Example of detecting protocols in global variables:
```
from pyrosetta.distributed.cluster import reproduce
from my_protocols import protocols # Protocols are in global variables

def main():
    reproduce(...)

if __name__ == "__main__":
    main()
```

Example of detecting protocols in local variables:
```
from pyrosetta.distributed.cluster import reproduce

def main():
    from my_protocols import protocols # Protocols are in local variables

    reproduce(...)

if __name__ == "__main__":
    main()
```


klimaj:
@lyskov Thanks for your review! Yes, that's a good question. The `reproduce` function allows the user to input the protocols as a list of functions or generators representing the original user-defined PyRosetta protocols, but they need to make sure the protocols are in the same order as in the original simulation. Because this information is already cached in the input PDB file or scorefile, it's intended to be a convenience for what is usually a second party reproducing the simulation where they have only checked out the GitHub repo with the protocol definitions and imported them into their script. As long as the original author of the simulation has not named their protocols the same as python builtin functions, then this auto-detection would work. Even on the off-chance of detecting an incorrect function, that function signature is very unlikely to match the function signature used by PyRosettaCluster (having a `PackedPose` object as the first argument and the special keyword arguments parsed by PyRosettaCluster), so the protocol will not run anyway unless it is the correct match. Albeit, I see the concern, and agree it's likely outside the scope of the PR since this is a bug fix for a functionality that has already been in PyRosettaCluster (but just not implemented correctly yet). I've also added explicit logging messages in this PR to indicate from which scope the functions have been detected, so this functionality is not used as blindly as before.

PR #485, opened 2025-07-26
Add the deterministic_flag option to the ProteinMPNNProbabilitiesMetric SimpleMetric

klimaj:
This PR adds the `deterministic_flag` option to the `ProteinMPNNProbabilitiesMetric` PerResidueProbabilitiesMetric. After some evidence provided in #429, the `ProteinMPNNMover` already has the `deterministic_flag` option set to `false` by default, and therefore due to Torch randomness the `ProteinMPNNProbabilitiesMetric` can return non-deterministic PSSM tables, even with the `-run:constant_seed 1` Rosetta option enabled. This PR just adds some control of the `ProteinMPNNMover` from the `ProteinMPNNProbabilitiesMetric` to enable deterministic mode when calculating per-residue ProteinMPNN probabilities.


MoritzErtelt (approved):
Looks good to me, thanks for adding the new option!

source/src/protocols/protein_mpnn/ProteinMPNNProbabilitiesMetric.cc
```diff
@@ -147,6 +148,8 @@ ProteinMPNNProbabilitiesMetric::parse_my_tag(
 		}
 	}
 
+	deterministic_flag_ = tag->getOption< bool >( "deterministic_flag", deterministic_flag_ );
```
roccomoretti, line 151: Is there a reason to name the user-facing option `deterministic_flag` versus something like simply `deterministic`?

klimaj: It's mainly geared to match the `ProteinMPNNMover` option, which called it `deterministic_flag`:
https://docs.rosettacommons.org/docs/latest/scripting_documentation/RosettaScripts/xsd/mover_ProteinMPNNMover_type

klimaj:
@MoritzErtelt @roccomoretti thanks for your reviews!

PR #503, opened 2025-08-18
Supporting a universal PyRosetta initialization file

klimaj:
This PR aims to add support for a PyRosetta initialization file type (i.e., a `.init` file) for reproducible PyRosetta initialization. 

What started out as a quick and dirty method to cache `.params` files morphed into a universal file type for streamlining PyRosetta initialization. This approach takes advantage of the `ProtocolSettingsMetric` SimpleMetric in PyRosetta to cache Rosetta command line options and parser script variables, and `zlib`, `base64`, and `json` libraries for compressing/decompressing arbitrary input text and binary files, including files within files (i.e., subfiles). Subfiles are cached by doing a brute force search through all input files by splitting file contents on spaces, which enables, for example, caching conformer files on `PDB_ROTAMERS` lines within `.params` files, and caching subfiles in list files passed in with the `-l` Rosetta command line flag. Note that the PyRosetta database and any input directories are not cached in the `.init` file.

The following new methods are supported in this PR:

- `pyrosetta.dump_init_file`: write a PyRosetta initialization `.init` file
- `pyrosetta.init_from_file`: initialize PyRosetta from a `.init` file
- `pyrosetta.get_init_options`: get the currently initialized Rosetta command line options
- `pyrosetta.get_init_options_from_file`: get PyRosetta initialization options from a `.init` file

This is a work in progress and suggestions/recommendations are welcome. Once merged, ideally the `.init` file format can remain stable over incremental PyRosetta versions.

PR #511, opened 2025-09-01
Integrating PyRosetta initialization files into PyRosettaCluster

klimaj:
Depends on #503. An accurate diff between this PR and #503 is here: https://github.com/klimaj/rosetta/compare/init...cluster_init

The purpose of this PR is to support several new features:
1. Adds an `output_init_file` instance attribute to `PyRosettaCluster`, enabling dumping of a `.init` or `.init.bz2` file upon instantiation.
2. Adds `author`/`email`/`license` instance attributes to `PyRosettaCluster`, which are cached in the `.init` or `.init.bz2` file and output decoy and scorefile metadata.
3. Enables the `input_file` keyword argument of the `pyrosetta.distributed.cluster.reproduce` method to accept a `.init` or `.init.bz2` file that initializes PyRosetta before simulation reproduction.
    - Also adds a `skip_corrections` keyword argument to enable skipping ScoreFunction corrections so that the reproduced results may be used for successive reproductions.
4. Adds a `pyrosetta.distributed.cluster.export_init_file` function that enables exporting an output decoy (in `.pdb`, `.pdb.bz2`, `.b64_pose`, `.b64_pose.bz2`, `.pkl_pose`, `.pkl_pose.bz2` format) to a `.init` or `.init.bz2` file format.
5. Adds a `norm_init_options` instance attribute to `PyRosettaCluster`, enabling normalization of the task's PyRosetta initialization options. This optional convenience feature takes advantage of the `pyrosetta.get_init_options` method to update the `options` and `extra_options` keyword arguments of each task after PyRosetta initialization in the `billiard` subprocess on the dask workers, which expands option names and relativizes any input files and directories to the `billiard` subprocess current working directory. Relativized paths are ideal for reproduction of simulations by a second party on a different filesystem.
6. Adds `pyrosetta.distributed.io.read_init_file` and `pyrosetta.distributed.io.init_from_file` functions, which handle `.init` and `.init.bz2` files.

Please note that this PR also depends on PR #462 supporting a `.b64_pose` and `.pkl_pose` file outputs in `PyRosettaCluster`. The impetus for supporting a `.init` file in the `PyRosettaCluster` simulation reproduction life cycle is that loading a `.b64_pose` or `.pkl_pose` file into memory requires that PyRosetta is initialized with the same residue type set as that used to save the `.b64_pose` or `.pkl_pose` file (otherwise PyRosetta doesn't know how to reconstruct the `Pose`, resulting in a segfault). Effectively, a user remains locked out of `.b64_pose` and `.pkl_pose` files unless PyRosetta is initialized correctly, which can be easily accomplished by PyRosetta initialization with a `.init` or `.init.bz2` file. Hence, if a user decides to output results in `.b64_pose` or `.pkl_pose` format, the `.init` file can then be used to initialize PyRosetta identically and load the `.b64_pose` or `.pkl_pose` file into memory.


klimaj:
@lyskov thanks, yes this PR adds an HMAC signature and a SHA256 signature to the `.init` file metadata that is signed by `PyRosettaCluster`, but doesn't affect any user-facing code. The goal of these signatures is to detect tampering with the PyRosetta `.init` file output by `PyRosettaCluster`, which ought to never be modified outside of any PyRosetta methods. Upon loading the `.init` file, the HMAC and SHA256 signatures are recomputed, and if they differ from that used to sign the original `.init` file, then we raise an exception with a helpful error message signaling that the `.init` file has been tampered with and specify the workaround (which is to delete the offending signature from the `.init` file metadata -- this is not recommended, but it still allows the `.init` file to be loaded). So this implementation validates the data integrity with a special PyRosettaCluster output `.init` file metadata format. I'm open to adding documentation somewhere. Generally, though, the `.init` file metadata is arbitrary text that any user can add when running `pyrosetta.dump_init_file`.

PR #512, opened 2025-09-02
Add cooldown time for PyRosettaCluster unit tests

klimaj:
This quick PR aims to add a `cooldown_time` instance attribute to `PyRosettaCluster` (with a default of 1 second) so loggers can flush after the simulation is complete but before unit test cases exit. This makes more sense than the unit test sleeping after already exiting.


lyskov (approved):
LGTM!

PR #521, opened 2025-09-18
Add warning for pose.cache security risks

rclune:
Added warning in the PoseCacheAccessor class about the security risks of using the pickle module.


lyskov (approved):
LGTM, - thank you @rclune !

PR #543, opened 2025-10-14
add PyRosetta example of using Python Mover in XML scripts

lyskov:


source/src/python/PyRosetta/src/test/T852_SubClassing_XML_Mover.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+
+# Create Custom Mover ----------------------------------------------------------------------------------------------------------
+from pyrosetta import get_fa_scorefxn
+from pyrosetta.rosetta import core, protocols, basic, utility
+from pyrosetta.rosetta.protocols.moves import Mover, MoverCreator, MoverFactory
+from pyrosetta.rosetta.utility.tag import XMLSchemaDefinition, xsct_positive_integer, XMLSchemaComplexTypeGenerator, XMLSchemaAttribute, XMLSchemaType
+from pyrosetta.rosetta.std import list_utility_tag_XMLSchemaAttribute_t
+from pyrosetta.rosetta.core.scoring import ScoreFunction, ScoreFunctionFactory, attributes_for_parse_score_function_w_description, parse_score_function
```
woodsh17: The imports from core.scoring are no longer needed since we don't read in a sfxn for the Mover, this was leftover from what I had before

source/src/python/PyRosetta/src/test/T852_SubClassing_XML_Mover.py
```diff
@@ -0,0 +1,173 @@
...
+    def create_mover(self):
+        mover = BootCampMoverPy()
+        self.instances_.append(mover)
+        return mover
+
+    def keyname(self):
+        return BootCampMoverPy.mover_name()
+
+    def provide_xml_schema(self, xsd):
+        print("creator provide_xml_schema is called")
```
woodsh17: Probably don't want these print statements anymore (also line 136).

source/src/python/PyRosetta/src/test/T852_SubClassing_XML_Mover.py
lyskov: Good catch! I removed most of it, but turned out that `attributes_for_parse_score_function_w_description` is needed.

PR #544, opened 2025-10-17
Test for beta scorefunction discrepancy with clone and minimization.

roccomoretti:
@Bloeci noticed a substantial scoring discrepancy when working with poses which had been cloned and then minimized. It looks like there's an issue when you score the pose before cloning and then go into minimization. -- Notably, this seems to only be the case for beta & beta_nov16 energy functions.

I've tracked this down to some issue with the Energy Graph. If you reset the energy graph (basically delete and then re-add the EnergyEdges) this discrepancy goes away. -- I can't understand why, though, because it goes away even if you put all the energy edges back the way they were. I'm also not sure how/why that would be interacting with the beta_nov16 specifically.

Lacking the ability to take this any further, I've encapsulated things into a unit test. Hopefully this will help someone to investigate further.

(P.S. It looks like only the errant score-before-cloned minimization is resulting in the "Inaccurate G" messages -- which may or may lead into the discrepancy.)

### DIFF
--- a/source/test/core.test.settings
+++ b/source/test/core.test.settings
@@ -361,6 +361,7 @@ sources = {
 		"ScoreFunction",
 		"ScoreFunctionFactory",
 		"ScoreTest",
+		"ScoreFunction_consistency",
 		"SplitUnfoldedTwoBodyPotential",
 		"SurfacePotential",
 		"SymmB3G",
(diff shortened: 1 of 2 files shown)


zznidar:
I think I've already asked Max this during the workshop, but does the minimiser take any of the already-calculated energies/forces/any other vectors? Because the shape of the minimised molecule was also different, as though it started with different starting conditions.

PR #548, opened 2025-10-22
Add beta_jan25 energy function

Haddox:
Hi @roccomoretti and @lyskov,

@fdimaio and I have developed a new beta energy function called `beta_jan25`. This is an updated version of `beta_nov16`.

The aim of this PR is to add `beta_jan25` to the `rosetta` source code.

We will shortly post a manuscript describing how we developed `beta_jan25`. They key updates are to the LJ potential. We identified steric clashing in proteins that were relaxed or designed using `beta_nov16`. We identified examples of this problem in a high-quality benchmark from the `dualoptE` protocol used to train the energy function. We then used this benchmark, and the others in `dualoptE`, to refit a small number of LJ parameters. The refitting largely eliminated the clashing problem, and `beta_jan25` is as good or better than `beta_nov16` when assessed on multiple benchmarks using validation data.

Please let me know if you have any comments or questions. I am also happy to share a draft of the manuscript. I am having Frank review my changes, and then I can ping one or both of you when Frank gives the thumbs up.

@fdimaio: could you please review my changes? And let me know if there are any additional changes I need to make? A few questions I had are:
- should genpot be "on top" of `beta_jan25`? Or should we stick with `beta_nov16`? See line 752 of `score_function_corrections.cc`
- currently, the `-beta` flag does not give the same results as either the `-beta_jan25` flag or the `-beta_nov16` flag. At least one reason for this is it seems to invoke the `beta_genpot.wts` file. Is this the correct behavior?

Thanks!

Hugh

### DIFF
--- a/source/src/python/PyRosetta/src/pyrosetta/distributed/cluster/__init__.py
+++ b/source/src/python/PyRosetta/src/pyrosetta/distributed/cluster/__init__.py
@@ -50,7 +50,7 @@ __all__: List[str] = [
     "run",
     "update_scores",
 ]
-__version__: str = "3.3.0"
+__version__: str = "3.4.0"
 
 
 with warnings.catch_warnings() and suppress(NameError):
--- a/source/src/python/PyRosetta/src/pyrosetta/distributed/cluster/base.py
+++ b/source/src/python/PyRosetta/src/pyrosetta/distributed/cluster/base.py
@@ -20,11 +20,13 @@ except ImportError:
     )
     raise
 
+import logging
 import pyrosetta
 import pyrosetta.distributed
 
 from datetime import datetime
 from functools import wraps
+from pyrosetta.distributed.cluster.config import __dask_version__
 from pyrosetta.distributed.cluster.converters import _parse_protocols
 from pyrosetta.distributed.cluster.initialization import (
     _get_norm_task_options,
@@ -33,6 +35,7 @@ from pyrosetta.distributed.cluster.initialization import (
 from pyrosetta.distributed.cluster.serialization import Serialization
 from pyrosetta.distributed.cluster.validators import (
     _validate_clients_indices,
+    _validate_priorities,
     _validate_protocols_seeds_decoy_ids,
     _validate_resources,
 )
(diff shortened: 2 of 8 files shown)


database/scoring/weights/beta_jan25.wts
```diff
@@ -0,0 +1,32 @@
+# beta_nov16
+#   beta energy function following parameter refitting (Frank DiMaio and Hahnbeom Park), November 2016
```
roccomoretti: Please update the comments at the top to be accurate

lyskov:
Looks good to me! @Haddox please let us know when this is ready to be merged. Thanks,

roccomoretti:
Looks like things need to be beautified. (`python ../tools/python_cc_reader/beautify_changed_files_in_branch.py` from the source/ directory)

Regarding the  ga_ligand results change, does the beta_jan25 scorefunction pull in the ligand torsional changes from gen_potential, or are those omitted? Is running ga_ligand_dock with beta_jan25 appropriate? -- If you do want to run ga_ligand_dock with gen_potential rather than beta_jan25, then it might be appropriate to switch the integration tests to use best practices. (It looks like they are mixing the `-gen_potential` command-line flag with the `beta.wts` weights in the XML. I'm not sure if that environment/weight mismatch is an issue.)

Haddox:
@roccomoretti: regarding the ga_ligand tests, Frank advised me to update the weights files to use genpot weights. I just pushed those updates. Thanks for catching that.

Also, it appears that my branch now has conflicts that must be resolved. I think this might relate to the commit (1314848) that I accidentally pulled in. Could you please advise me on the best way to address this conflict?

Thanks!

roccomoretti:
You should be able to use the "Resolve conflicts" button on the Github website to change things. (Just select the contents of main)

Alternatively, just merge your branch with `main` (`git fetch`, `git merge origin/main`) and the resolve the conflicts locally (for files you shouldn't be changing with your branch, you can just do `git checkout origin/main -- <filename>`).

We use squash merges, so you don't need to worry about clean history (everything is squashed to a single commit anyway).

PR #551, opened 2025-10-27
Support chain designations of multiple letters.

roccomoretti:
Part of the "better support for mmCIF format" effort.

The mmCIF format supports larger systems with multi-character chain designations. This PR changes the internals of Rosetta to support multi-letter chain designations. There shouldn't be any significant behavioral changes if you just continue to use single letter chain designations.

The most significant interface change was the change in interface designation. Instead of strings like "A_B", code now generally takes a `core::pose::DockingPartners` object. (This can be created from the string designation, to ease IO - it should be invisible except perhaps at the PyRosetta level.)

Places that needed to be fixed were discovered primarily with type change issues during compilation. I may have missed some places which need updated support. 

Apologies for the massive size of this PR, it would be hard to make such a fundamental change in any smaller unit.


lyskov (approved):
Tentatively approving, pending successful test runs.

roccomoretti:
Bunch of cosmetic tracer output changes. The MEM residue gets a different chain designation in some cases. Otherwise tests look clean.

PR #556, opened 2025-10-29
Fixes for compiling with gcc15

roccomoretti:
Assorted fixes for compiling with GCC 15. Should be combined with #555


roccomoretti:
All of the relevant changes have already been manually merged with #555, rendering this an effective no-op.

PR #655, opened 2026-04-13
Quick update: change which HKDF function is used by HMAC-SHA256 verification machinery in PyRosettaCluster

klimaj:
This quick patch has virtually no effect on security or runtime functionality, it just updates which extract-and-expand hash-based message authentication code (HMAC)-based key derivation function (HKDF) is used for pseudorandom key generation in task dictionary HMAC-SHA256 verification, reflecting original intention.

To be merged after #648 for incremental versioning.


lyskov (approved):
👍

PR #659, opened 2026-04-27
Fix memory leaks in Inline_File_Provider

lyskov-ai:
## Summary

- `clear_input_files()` and `remove_input_file()` were leaking `std::stringstream` objects by calling `vector::clear()`/`erase()` without deleting the raw pointers
- `get_ostream()` also leaked a `new std::stringstream` on every call
- Fixed by switching `input_files` and `output_files` members from `std::vector<std::pair<std::string, std::stringstream*>>` to `std::vector<std::pair<std::string, std::unique_ptr<std::stringstream>>>`, so the vector destructor and `erase()` handle cleanup automatically

## Test plan

- [ ] Build `utility` library in debug mode: `cd source && ./scons.py -j16 mode=debug utility`
- [ ] Run utility unit tests: `cd source && python test/run.py -j16 --mode=debug`

### DIFF
--- a/source/src/utility/inline_file_provider.cc
+++ b/source/src/utility/inline_file_provider.cc
@@ -30,10 +30,9 @@ void Inline_File_Provider::init_static_inputs(){
 }
 
 void Inline_File_Provider::add_input_file( const std::string &filename, const std::string &contents ){
-	auto *newstream = new std::stringstream(contents);
 	std::string filtered_filename( standardise_filename( filename ) );
 	std::cout << "Creating inline input file: '" << filename << "' = (" << filtered_filename << ")" << std::endl;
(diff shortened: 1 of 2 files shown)


roccomoretti (approved):
Looks good, pending tests.

Was this a bug that was actually found in the wild, or was the dangling pointer issue just found from code review? (I ask because I was under the impression the inline file provider wasn't really being used anymore.)

lyskov:
> Looks good, pending tests.
> 
> Was this a bug that was actually found in the wild, or was the dangling pointer issue just found from code review? (I ask because I was under the impression the inline file provider wasn't really being used anymore.)

This was found by doing only code review. Since there was no other changes in the code I assume the code in question is indeed no longer used.

lyskov (approved):

PR #665, opened 2026-04-28
Explicitly delete copy constructor and assignment in Timer

lyskov-ai:
## Summary
`Timer::~Timer()` calls `Report()`, logging the elapsed time to a tracer or stdout. With no explicit copy policy, a copied `Timer` would produce a second `Report()` on its own destruction — silently logging a duplicate timing entry for the same measured interval.

`tag_` and `t_` are non-owning pointers (no `delete` in the destructor), so there is no memory-safety risk. The issue is purely semantic, but it is a real hazard for callers who pass or store `Timer` by value.

Explicitly `= delete` the copy constructor and copy assignment to make the non-copyable intent clear.

## Test plan
- [x] `basic/gpu` builds cleanly in debug mode
- [x] Full debug build passes with no errors

### DIFF
--- a/source/src/basic/gpu/Timer.hh
+++ b/source/src/basic/gpu/Timer.hh
@@ -39,6 +39,8 @@ public:
 	Timer();
 	~Timer();
 
+	// ~Timer() calls Report(), so a copy would fire a second Report() on its
+	// own destruction, silently logging a duplicate timing entry.
 	Timer(Timer const &) = delete;
 	Timer & operator=(Timer const &) = delete;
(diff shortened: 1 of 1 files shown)


lyskov (approved):

roccomoretti (approved):
Probably want to leave a comment in the code as to why they're being deleted, as I don't know if it will be obvious.

lyskov:
Good point @roccomoretti ! I added it and will see if we can get more comments in the future.

lyskov (approved):

PR #666, opened 2026-04-28
Hotfix of T919 and T920 tests for ubuntu.clang.cxx11thread.serialization.python38.PyRosetta.unit test

klimaj:
This PR removes a few built-in `hash()` function calls from the `T919` and `T920` unit tests, which are non-deterministic in Python-3.8.


lyskov (approved):
LGTM, - thank you @klimaj !

PR #682, opened 2026-05-01
Beta barrel parametrization

everyday847:


vmullig:
Note that the helical bundle machinery does permit beta-barrels and TIM
barrels to be parametrically generated as well.  See the beta strand
parameters (strands are special cases of helices), and the
MakeBundle/PerturbBundle/BundleGridSampler documentation.

everyday847:
That's true. Really maybe the right approach is to think of this as a convenience thing: each more specialized class could ensure that the parameters being requested are generally in the range of helices/strands in their internal and collective parameters. What do you think?

vmullig:
Yes, makes sense.  I'm not really doing any more Rosetta development, so
your call!

PR #683, opened 2026-05-01
Add correlated parameter perturbation to PerturbBundle

everyday847:
Introduces <Correlation> sub-tags for PerturbBundle that specify Pearson correlation coefficients between bundle parameters. When present, correlated parameters are perturbed jointly via multivariate Gaussian sampling (Cholesky decomposition of the covariance matrix). Parameters not involved in correlations continue to be perturbed independently.

Adds a Cholesky decomposition wrapper in numeric/linear_algebra/ backed by Eigen::LLT, exposing only vector1-based types.


vmullig:
Nice -- this is a feature that had been on my wishlist for a long time.

PR #698, opened 2026-05-11
Rework chirality detection in restype determination for PDB loading

roccomoretti:
A side quest from PR #687 

Currently, we have a rather brittle atom-name-based approach to chirality detection on PDB read-in, which isn't as robust as it could be, in large part due to trying to decide chirality early on. Instead of imposing L/D identity early, we can defer it until later with the atom name matching heuristics. This way, the D & L version of amino acids can co-exist with non-aa residues with the same three letter code (e.g. when we're loading a CCD type with the same three letter code as an unrelated Rosetta residue type), and the ResidueTypeFinder should pick the best match. 

I've also reworked the atom name matching heuristics, implementing the previously proposed idea to separate out virtual atom counts from non-virtual counts.


roccomoretti:
Aside from cosmetic changes in debug-level tracers, the  swm_dna_bridge and rna_denovo_dna_bridge integration tests change as the input PDBs type the first nucleotide in the chain as `CYT:LowerDNA:VirtualDNAPhosphate:Virtual_Phosphate` versus `CYT:LowerDNA:VirtualDNAPhosphate` due to the missing coordinates for `O5'`. This is arguably a more correct typing, as `CYT:LowerDNA:VirtualDNAPhosphate:Virtual_Phosphate` doesn't have that oxygen present as a real atom.

PR #706, opened 2026-05-15
Apply Rule of Zero to empty destructors across core/

lyskov-ai:
## Summary

Remove trivially empty (`~X() {}`) or `= default;`-bodied destructors across `core/`, **deleting the declaration outright** rather than re-declaring it as `= default` in the header. The now-redundant out-of-line `.cc` definitions are removed as well.

Per @roccomoretti's review, two corrections to the original approach:

1. **`~X() override = default;` is unnecessary.** For every derived class here the base already declares a virtual destructor, so the destructor is virtual by inheritance whether or not it is declared — the explicit declaration bought nothing. Omitting it is the actual Rule of Zero: it keeps the destructor virtual (inherited) and drops the boilerplate.
2. **Defaulting the destructor does *not* restore the implicit move members.** A `~X() = default;` destructor is still *user-declared*, which suppresses the implicit move constructor/assignment exactly as a user-provided one would. (An earlier version of this description wrongly claimed defaulting "unlocks" moves.) Only omitting the destructor entirely lets the implicit moves be generated — and then only for classes with no other user-declared special members.

Two classes are polymorphic bases that need an explicitly-declared virtual destructor and therefore **keep `virtual ~X() = default;`**: `LKHBondInfo` and `HBondInfo`.

## Affected sibling groups

- `core/conformation/membrane/{AqueousPoreParameters, ImplicitLipidInfo, MembraneGeometry, MembraneInfo}` and `membrane_geometry/{Bicelle, DoubleVesicle, Slab, Vesicle}` — 8 classes
- `core/conformation/parametric/{Boolean, Real, RealVector, Size, SizeVector}ValuedParameter` — 5 classes
- `core/io/silent/SilentFileData::{iterator, const_iterator}` — inline empty dtors removed
- `core/pack/interaction_graph/RotamerDots.{hh,cc}` — `DotSphere`, `RotamerDots`, `RotamerDotsCache`, `InvRotamerDots`
- `core/scoring/etable/EtableEnergy.{hh,cc}` — `EtableEvaluator`, `AnalyticEtableEvaluator`, `TableLookupEvaluator`
- `core/scoring/hbonds/graph/HBondInfo.hh` — `LKHBondInfo`, `HBondInfo` kept as `virtual ~X() = default;` (polymorphic roots); also removed an unused user-defined empty copy ctor on `LKHBondInfo`
- `core/scoring/nmr/NMRDummySpinlabelVoxelGrid.{hh,cc}` — `VoxelGridPoint`, `NMRDummySpinlabelAtom`, `VoxelGridPoint_AA`, `NMRDummySpinlabelVoxelGrid`
- `core/scoring/sc/MolecularSurfaceCalculator::Atom`
- `core/energy_methods/SAXSEnergy`

Verified that no `.cc` retains an out-of-line definition for any removed destructor, and the native Rosetta beautifier reports no formatting changes on the touched headers. Pure code-style refactor; no intended behavior change.

### DIFF
--- a/source/src/core/conformation/membrane/AqueousPoreParameters.cc
+++ b/source/src/core/conformation/membrane/AqueousPoreParameters.cc
@@ -99,8 +99,6 @@ AqueousPoreParameters::AqueousPoreParameters(
 	pore_rotation_angle_( pore_rotation_angle )
 {}
 
-AqueousPoreParameters::~AqueousPoreParameters() {}
-
 AqueousPoreParameters::AqueousPoreParameters( AqueousPoreParameters const & src )
 : utility::VirtualBase( src ),
 	min_center_x_( src.min_center_x_ ),
--- a/source/src/core/conformation/membrane/AqueousPoreParameters.hh
+++ b/source/src/core/conformation/membrane/AqueousPoreParameters.hh
@@ -70,7 +70,6 @@ public:
 	);
 
 	AqueousPoreParameters(AqueousPoreParameters const & src);
-	~AqueousPoreParameters() override;
 
 	AqueousPoreParametersOP
 	clone() const;
--- a/source/src/core/conformation/membrane/ImplicitLipidInfo.cc
+++ b/source/src/core/conformation/membrane/ImplicitLipidInfo.cc
@@ -98,8 +98,6 @@ ImplicitLipidInfo::ImplicitLipidInfo(
 	initialize_implicit_lipid_electricfield_parameters();
 }
 
-ImplicitLipidInfo::~ImplicitLipidInfo() {}
-
 ImplicitLipidInfoOP
 ImplicitLipidInfo::clone() const {
 	return ImplicitLipidInfoOP( new ImplicitLipidInfo( *this ) );
--- a/source/src/core/conformation/membrane/ImplicitLipidInfo.hh
+++ b/source/src/core/conformation/membrane/ImplicitLipidInfo.hh
@@ -62,9 +62,6 @@ public:
 		core::Real temperature
 	);
 
-
-	~ImplicitLipidInfo() override;
-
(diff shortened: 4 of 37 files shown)


lyskov (approved):

lyskov (approved):

roccomoretti (approved):
Fine as far as it goes, but putting `~Class() override = default` is unnecessary, as the only benefit of declaring a default destructor is being able to label it virtual. (Which it already has to be with the `override`.)

IIRC, the "unlocks the implicit move-special-members" isn't quite correct -- explicitly defaulting the destructor still counts as "user defined" for the purposes of suppressing the implicit move functions. (Which is why just leaving it off is typically preferred.)

lyskov-ai:
Thanks @roccomoretti — you're right on both counts, and I've updated the PR accordingly.

- **Removed the `~X() override = default;` declarations entirely** (23 classes) instead of defaulting them. Since each derives from a base with a virtual destructor, the destructor stays virtual by inheritance, so the declaration was redundant. Did the same for the three non-virtual `~X() = default;` (`DotSphere`, `RotamerDotsCache`, `MolecularSurfaceCalculator::Atom`).
- **Kept `virtual ~X() = default;` only on the two polymorphic roots** that need an explicitly-declared virtual destructor: `LKHBondInfo` and `HBondInfo`.
- **Corrected the description**: as you noted, a `= default` destructor is still user-declared and does not un-suppress the implicit move members — that only happens when the destructor is omitted entirely, which is now what this does (and only actually takes effect for the classes with no other user-declared special members).

Verified no `.cc` still defines a removed destructor out-of-line, and the beautifier reports no formatting changes on the touched headers. (A full debug build is currently blocked on my end by the unrelated GCC 15 `numeric/` template-id break — happy to run it once that lands.)

PR #708, opened 2026-05-16
Testing Pose bindings/numeric unit tests with standard PyRosetta builds

klimaj:
This PR splits out two tests from `T900_distributed.py` that don't require the `pyrosetta.distributed` framework, but instead only rely on `numpy` and ought to be tested by the standard PyRosetta builds.

### DIFF
--- a/source/src/python/PyRosetta/src/test/T021_Pose_Methods.py
+++ b/source/src/python/PyRosetta/src/test/T021_Pose_Methods.py
@@ -1,0 +1,13 @@
+# :noTabs=true:
+# (c) Copyright Rosetta Commons Member Institutions.
+# (c) This file is part of the Rosetta software suite and is made available under license.
+# (c) The Rosetta software is developed by the contributing members of the Rosetta Commons.
+# (c) For more information, see http://www.rosettacommons.org.
+# (c) Questions about this can be addressed to University of Washington CoMotion, email: license@uw.edu.
+
+__author__ = "Jason C. Klima"
+
+from utils.distributed import run_unittest
+
+if __name__ == "__main__":
+    run_unittest("pyrosetta.tests.bindings.core.test_pose", timeout=60)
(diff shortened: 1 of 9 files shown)


source/src/python/PyRosetta/src/test/T021_Pose_Methods.py
```diff
@@ -0,0 +1,14 @@
...
+# (c) For more information, see http://www.rosettacommons.org.
+# (c) Questions about this can be addressed to University of Washington CoMotion, email: license@uw.edu.
+
+__author__ = "Jason C. Klima"
+
+from utils.distributed import exit_if_missing_numpy_requirement, run_unittest
+
+if __name__ == "__main__":
+    exit_if_missing_numpy_requirement()
+    run_unittest("pyrosetta.tests.bindings.core.test_pose", timeout=60)
```
lyskov, line 13: -- i guess this is one way to take care of this issue but the issue is that test will get silently "green" if no numpy was installed. Have you considered updating `get_required_pyrosetta_python_packages_for_release_package` so it returns at least `numpy` for non-distributed platform? That way we should always be able to run this test. Thoughts?

https://github.com/RosettaCommons/rosetta/blob/main/tests/benchmark/tests/__init__.py#L520

klimaj: @lyskov Thanks for reviewing, and good point - I have now updated both `get_required_pyrosetta_python_packages_for_release_package` and `get_required_pyrosetta_python_packages_for_testing` to directly install `numpy` and removed the silent green mechanism (for reference, this is the same mechanism we currently use to skip T900-T921 tests on non-serialization builds, which are not fully silent since they print warnings when skipping due to missing packages). I agree, let's not exacerbate the issue by adding more skipping behavior. The T021 and T022 tests now pass without being skipped and will error out if `numpy` is missing.

lyskov (approved):

source/src/python/PyRosetta/src/test/T021_Pose_Methods.py
lyskov: Hm… the "skip" in T900-T921 is quite different: package still useful without the extra dependencies while in this case it is not.

PR #720, opened 2026-06-24
Fix restart behavior in PyJobDistributor

### LINKED ISSUE #347: Infinite while loop in PyJobDistributor
There is a bug in the PyRosetta [PyJobDistributor](https://github.com/RosettaCommons/rosetta/blob/main/source/src/python/PyRosetta/src/pyrosetta/toolbox/py_jobdistributor.py) method start_decoy:

```
    def start_decoy(self):
        while self.sequence:

            self.current_id = self.sequence[0]
            self.current_name = self.pdb_name + "_" + str(self.current_id) + (".pdb.gz" if self.compress else ".pdb")
            self.current_in_progress_name = self.current_name + ".in_progress"

            if (not os.path.isfile(self.current_name)) and (not os.path.isfile(self.current_in_progress_name)):
                with open(self.current_in_progress_name, "w") as f:
                    f.write("This decoy is in progress.")
                print("Working on decoy: {}".format(self.current_name))
                break
```

If the current file in sequence is complete or in progress, in the case of multiple processes working on the same job in parallel, the while loop stalls on this decoy. Adding something like this else block after will solve the issue.

```
    def start_decoy(self):
        while self.sequence:

            self.current_id = self.sequence[0]
            self.current_name = self.pdb_name + "_" + str(self.current_id) + (".pdb.gz" if self.compress else ".pdb")
            self.current_in_progress_name = self.current_name + ".in_progress"

            if (not os.path.isfile(self.current_name)) and (not os.path.isfile(self.current_in_progress_name)):
                with open(self.current_in_progress_name, "w") as f:
                    f.write("This decoy is in progress.")
                print("Working on decoy: {}".format(self.current_name))
                break
            else:
                self.sequence.remove(self.current_id)
```

roccomoretti:
The PyJobDistributor can result in an infinite loop if one of the entries are already completed (e.g. from a prior run) or is in progress from another run.

Fixes #347

### DIFF
--- a/source/src/python/PyRosetta/src/pyrosetta/toolbox/py_jobdistributor.py
+++ b/source/src/python/PyRosetta/src/pyrosetta/toolbox/py_jobdistributor.py
@@ -109,6 +109,9 @@ class PyJobDistributor:
                     f.write("This decoy is in progress.")
                 print("Working on decoy: {}".format(self.current_name))
                 break
+            else:
+                print("Skipping decoy {} as it's already completed or is otherwise in-progess.".format(self.current_name))
+                self.sequence.remove(self.current_id)
 
     def output_decoy(self, pose):
         if os.path.isfile(self.current_name): # decoy already exists, probably written to other process -> moving to next decoy if any


lyskov (approved):

lyskov:
Looks good to me, - thank you for fixing this @roccomoretti !

PR #724, opened 2026-06-26
Apply Rule of Zero to remaining energy-container neighbor iterators

lyskov-ai:
## Summary

Applies the Rule of Zero to the neighbor-iterator classes of the three remaining `LREnergyContainer`-derived energy containers — `DenseEnergyContainer`, `OneToAllEnergyContainer`, and `PolymerBondedEnergyContainer` — completing the series begun in #689 (disulfide energy iterators) and #699 (remaining `LREnergyContainer` iterators). For all six iterator classes (`*NeighborIterator` and `*NeighborConstIterator`):

- **Remove the user-declared, defaulted destructor** (`~Foo() override;` in the header and `Foo::~Foo() = default;` in the `.cc`). The base classes `ResidueNeighborIterator` / `ResidueNeighborConstIterator` (both deriving from `utility::VirtualBase`) declare a virtual destructor, so the implicitly-generated destructor is virtual and does the same job.

- **Modernize the old-style C++03 copy-assignment prevention** — a private, declared-but-undefined derived-type `operator=` — to a public `= delete`. Behavior is unchanged: derived-to-derived assignment remains forbidden, so all assignment continues to flow through the polymorphic `operator=(BaseType const &)` override that downcasts and copies the members. A short comment records why the `= delete` is present.

The iterators hold only non-owning back-pointers into their container, so they never owned anything in their destructors; this is a pure ownership/semantics cleanup with no behavioral change.

## Note on diff size

At ~49 lines this is below the usual bundling threshold, but it is the complete remaining set of this exact sibling pattern: with the disulfide (#689) and `LREnergyContainer` (#699) iterators already done, `DenseEnergyContainer`, `OneToAllEnergyContainer`, and `PolymerBondedEnergyContainer` are the only `LREnergyContainer`-derived containers left with the private-undefined-`operator=` + defaulted-virtual-destructor idiom. There are no further same-pattern siblings to bundle.

### DIFF
--- a/source/src/core/scoring/DenseEnergyContainer.cc
+++ b/source/src/core/scoring/DenseEnergyContainer.cc
@@ -33,8 +33,6 @@ namespace scoring {
 
 ///////////////////////////////////////////////////////
 
-DenseNeighborIterator::~DenseNeighborIterator()= default;
-
 DenseNeighborIterator::DenseNeighborIterator(
 	Size const pos1_in,
 	Size const pos2_in,
@@ -147,8 +145,6 @@ DenseNeighborIterator::energy_computed() const
 }
 
 /////////////////////////////////////////////////////
-
-DenseNeighborConstIterator::~DenseNeighborConstIterator()= default;
 
 DenseNeighborConstIterator::DenseNeighborConstIterator(
 	Size const pos1_in,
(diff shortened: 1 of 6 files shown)


lyskov-ai:
Superseded by #699, which merged the identical Rule-of-Zero fix to DenseEnergyContainer, OneToAllEnergyContainer, and PolymerBondedEnergyContainer (and additionally covers ConstraintEnergyContainer, which this PR did not touch). Closing as redundant.

PR #726, opened 2026-06-26
Update deprecated Dask `LocalCluster` method parameters in PyRosetta unit tests

klimaj:
This quick PR aims to update some PyRosetta tests' usage of the deprecated `diagnostics_port` parameter in Dask's `LocalCluster` constructor, resulting in the following error since release `2026.6.0`:
```
TypeError: Server.__init__() got an unexpected keyword argument 'diagnostics_port'
```
Reference: https://b3.graylab.jhu.edu/test/913839

### DIFF
--- a/source/src/python/PyRosetta/src/pyrosetta/tests/distributed/test_dask.py
+++ b/source/src/python/PyRosetta/src/pyrosetta/tests/distributed/test_dask.py
@@ -32,21 +32,22 @@ class TestDaskDistribution(unittest.TestCase):
             if not self._dask_scheduler:
                 n_workers = 2
                 threads_per_worker = 2
-                diagnostics_port = None
+                dashboard_address = None
+                local_cluster_kwargs = dict(
+                    n_workers=n_workers,
+                    threads_per_worker=threads_per_worker,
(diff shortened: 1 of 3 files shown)


lyskov (approved):
Looks good to me (pending testing), - thank you for taking care of this @klimaj ! 
 
- have you considered also fixing the version of dependency (distributed package?) - maybe it will make it more future proof. 

Thanks,

klimaj:
@lyskov thanks for reviewing - we could fix the `distributed` package version, but we also need to support Python 3.9-3.14 versions on the Benchmark server, which at the moment dynamically select their most recent compatible `distributed` versions (for example, the `ubuntu-20.04.clang.cxx11thread.serialization.python39.PyRosetta.unit` test still selects `distributed==2024.8.0` which supports the `diagnostics_port` parameter). Furthermore, as tests may be run outside of the Benchmark server with various environments, it seems a bit more apt to patch the test itself, rather than the Benchmark test harness. This also allows catching deprecation errors with tests before scripts in the wild, hopefully allowing updating the PyRosetta source code if needed.

PR #707, opened 2026-05-15
Fix off-by-one loops over 1-indexed residue/atom/element ranges

lyskov-ai:
## Summary

Audit-pass fix for a family of off-by-one bugs of the same shape across 35 files. Continues the pattern from #695 and #701; this batch covers loops the earlier audits did not reach.

### Off-by-one in 1-indexed loops (33 files)

Each loop uses `for ( Size i = 1; i < container.size(); ++i )` (or `< pose.size()` / `< pose.total_residue()` / `< natoms()` / `< nres`) and the body uses `i` as a direct 1-indexed element accessor — `pose.residue(i)`, `vec[i]`, `xyz(i)`, etc. The last element is therefore silently skipped. Replaced with `<=` to include it.

Affected functions cover residue iteration in:
- `core/energy_methods/RNA_LJ_BaseEnergy.cc` (RNA neighbor scan)
- `core/io/pose_to_sfr/PoseToStructFileRepConverter.cc` (SS element enumeration — a 1-residue SS element at the C-terminus was dropped)
- `core/pack/interaction_graph/SurfacePotential.cc` ×3 (neighbor count, surface energy sum)
- `core/pose/rna/util.cc` (heavy-atom contact scan)
- `core/select/residue_selector/JumpUpstreamSelector.cc` (subset assignment, divergent from sibling `JumpDownstreamSelector` which uses `<=`)
- `protocols/antibody/residue_selector/CDRResidueSelector.cc`
- `protocols/canonical_sampling/mc_convergence_checks/HierarchicalLevel.cc` (address match count)
- `protocols/cartesian/md.cc` ×2 (per-atom state save/derivative check)
- `protocols/cutoutdomain/CutOutDomain.cc` (find_nearest_res)
- `protocols/denovo_design/components/FoldGraph.cc`
- `protocols/docking/DockingEnsemblePrepackProtocol.cc` ×2 (chain identity validation)
- `protocols/docking/metrics.cc` ×2 (cutpoint scan in Irmsd / contact metrics)
- `protocols/enzdes/EnzRepackMinimize.cc` (movable-residue collection for backrub)
- `protocols/features/strand_assembly/StrandBundleFeatures.cc` ×2 (per-atom distance scan)
- `protocols/fldsgn/MatchResidues.cc` ×2 (RMSD/superimpose atom-id map build)
- `protocols/frag_picker/GrabAllCollector.hh` (`clear`)
- `protocols/frag_picker/VallProvider.cc` (`find_chunk`)
[...]

PR #730, opened 2026-07-02
Fix off-by-one in 1-indexed loops in core/ (split 1/4 of #707)

lyskov-ai:
## Summary

One of **four PRs splitting #707** (originally a single 35-file off-by-one batch) into coherent, per-subsystem pieces, so each can be reviewed and its regression-test impact assessed independently. #707 is being closed in favor of these four. The split is deliberate — the individual diffs are below the usual bundling threshold — because the changes are behavior-affecting and the maintainer asked to isolate regression-test impact per subsystem.

**This PR: `core/` (5 files).**

Each loop uses `for ( Size i = 1; i < container.size(); ++i )` (or `< pose.size()` / `< nheavyatoms()` / `< upstream.size()`) and the body uses `i` as a direct 1-indexed accessor — `pose.residue(i)`, `vec[i]`, `xyz(i)` — so the last element is silently skipped. Changed `<` to `<=` to include it.

- `core/energy_methods/RNA_LJ_BaseEnergy.cc` — RNA neighbor scan
- `core/io/pose_to_sfr/PoseToStructFileRepConverter.cc` — SS-element enumeration; a 1-residue SS element at the C-terminus was dropped
- `core/pack/interaction_graph/SurfacePotential.cc` ×3 — neighbor count, surface-energy sum
- `core/pose/rna/util.cc` — heavy-atom contact scan
- `core/select/residue_selector/JumpUpstreamSelector.cc` — subset assignment; diverged from sibling `JumpDownstreamSelector`, which already uses `<=`

## Regression tests

None of the changed regression tests reported for #707 (`enzdes`, `inverse_rotamer_remodel`, `ligand_dock_cholesterol`, `mp_f19_relax`, `mpil_load_implicit_lipids`, `pose_sewing`, `unfolded_state_energy_calc`) fall in this subsystem, so this PR is expected to be inert with respect to that set. CI will confirm.

### DIFF
--- a/source/src/core/energy_methods/RNA_LJ_BaseEnergy.cc
+++ b/source/src/core/energy_methods/RNA_LJ_BaseEnergy.cc
@@ -325,7 +325,7 @@ RNA_LJ_BaseEnergy::eval_atom_energy(
 
 	Vector const heavy_atom_i( rsd1.xyz( m ) );
 
-	for ( Size j = 1; j < pose.size(); j ++ ) {
+	for ( Size j = 1; j <= pose.size(); j ++ ) {
 		if ( i == j ) continue;
 
 		conformation::Residue const & rsd2( pose.residue( j ) );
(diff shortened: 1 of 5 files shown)


roccomoretti (approved):
Looks reasonable, pending an update to the SurfacePotential unit tests.

PR #732, opened 2026-07-02
Fix off-by-one loops + self-distance bug in SEWING/splice/de novo/RNA (split 3/4 of #707)

lyskov-ai:
## Summary

One of **four PRs splitting #707** (originally a single 35-file off-by-one batch) into coherent, per-subsystem pieces, so each can be reviewed and its regression-test impact assessed independently. #707 is being closed in favor of these four. The split is deliberate — the individual diffs are below the usual bundling threshold — because the changes are behavior-affecting and the maintainer asked to isolate regression-test impact per subsystem.

**This PR: SEWING / splice / de novo / RNA structural assembly (10 files).** Two related fixes.

### Off-by-one in 1-indexed loops (`<` → `<=`)

The last element was silently skipped:

- `legacy_sewing/conformation/Assembly.cc` ×2 — `pose_loop_anchors`, `disconnected_segments`; the last segment, which by definition has no next, was skipped — exactly the case the predicate `!has_next(i)` is meant to catch
- `pose_sewing/movers/BlockwiseAnalysisMover.cc`
- `pose_sewing/movers/OmnibusDisulfideAnalysisLabelerMover.cc`
- `splice/SampleRotamersFromPDB.cc` ×2 — rotamer-deduplication scan
- `splice/SpliceManager.cc` — PSSM row check ("go over all the PSSM segments")
- `stepwise/legacy/modeler/rna/StepWiseRNA_WorkingParametersSetup.cc`
- `rna/movers/RNAIdealizeMover.cc` — `CoordinateConstraint` application
- `denovo_design/components/FoldGraph.cc`
- `fldsgn/MatchResidues.cc` ×2 — RMSD / superimpose atom-id map build
- `features/strand_assembly/StrandBundleFeatures.cc` ×2 — per-atom distance scan

### Self-distance copy/paste bug (in the two `pose_sewing` movers)

`pose.residue(upstream_res).xyz(2).distance(pose.residue(upstream_res).xyz(2))` is identically `0`, so the `<= crit_dist_` gate was always satisfied and the distance filter was a no-op. Corrected the second operand to `downstream_res`.

## Regression tests

This PR touches **`pose_sewing`**. The output change is intentional (combined off-by-one + self-distance correction) but results-affecting and warrants scientific sign-off before merge. The other files here are not in the reported changed-test set.

### DIFF
--- a/source/src/protocols/denovo_design/components/FoldGraph.cc
+++ b/source/src/protocols/denovo_design/components/FoldGraph.cc
@@ -852,7 +852,7 @@ FoldGraph::compute_best_solution( SegmentNames const & staple_loops ) const
 						solutions[sol].push_back( new_visited );
 					} else {
 						Solution tmpset;
-						for ( core::Size t=1; t<solutions[sol].size(); ++t ) {
+						for ( core::Size t=1; t<=solutions[sol].size(); ++t ) {
 							tmpset.push_back( solutions[sol][t] );
 						}
 						tmpset.push_back( new_visited );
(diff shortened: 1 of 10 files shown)


roccomoretti (approved):
Looks fine, pending updates to the OmnibusDisulfideAnalysisLabelerMover unit test.

PR #749, opened 2026-08-12
Add IAM::set_interface(string) for PyRosetta.

roccomoretti:
BindCraft (and others) use the InterfaceAnalyzerMover.set_interface() function with a string. This was removed during the multi-character chain designation update.

This PR puts it back for PyRosetta.


lyskov (approved):
LGTM!

PR #759, opened 2026-08-27
Fix compliation for Clang 23 and GCC 16

roccomoretti:
Update for newer compilers. Mostly assorted warnings-as-errors fixes, including a few false-positives.


lyskov (approved):

